PR #8, opened 2025-11-06
feat(ui): add responsive homepage layout with light/dark mode support

sabri414:
Changes:

- [ ] Added responsive homepage in app/page.tsx
- [ ] created FeatureBlock.tsx component for modular feature sections 
- [ ] imported feature screenshots for both modes 
- [ ] Updated providers.tsx and layout.tsx for consistent theming 
- [ ] verified compatibility with NextUI + TailwindCSS

Future Enhancements
Add subtle fade-in animations (Framer Motion)
Add hero CTA button for sign-up/login
Improve accessibility labels for images

Checklist
Homepage layout matches design mockup
Light/Dark mode switches images correctly
Tested responsiveness on mobile and desktop
Builds and runs locally without errors

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -16,7 +16,7 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-import { danger, fail, warn } from 'danger';
+import { danger, fail, warn, message } from 'danger';
 
 // Regular expression for conventional commits
 // UPDATE CONTRIBUTING.md PR TYPE LIST WHEN MAKING ANY REGEX CHANGES
@@ -62,6 +62,11 @@ const deletedMigrations = danger.git.deleted_files.filter(f =>
 const modifiedMigrations = danger.git.modified_files.filter(f =>
   f.startsWith('prisma/migrations/')
 );
+
+if (modifiedSchema)
+  message(
+    '`prisma/schema.prisma` was modified. Is the [dbdiagram.io schema](<https://dbdiagram.io/d/Tasktix-691fa53e228c5bbc1ad3060f>) up to date?'
+  );
 
 if (modifiedSchema && addedMigrations.length === 0) {
   fail(
--- a/docs/api-schema.yaml
+++ b/docs/api-schema.yaml
@@ -1,0 +1,759 @@
+openapi: 3.0.0
+info:
+  description: A Task Tracking Tool for Power Users
+  version: '1.0.0'
+  title: Tasktix API
+  license:
+    name: AGPL 3.0
+    url: 'https://www.gnu.org/licenses/agpl-3.0.txt'
+paths:
+  /user:
+    post:
+      summary: Create a new user account
+      tags:
+        - user
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/UserAccount'
+                - type: object
+                - required: [username, email, password]
+        required: true
+      responses:
+        '201':
+          description: User successfully created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreationResponse'
+              example:
+                message: User created
+                content: /api/user/{userid}
+        '400':
+          description: Invalid request body
+        '500':
+          description: Internal server error
+  /user/{id}:
+    patch:
+      summary: Update a user account
+      tags:
+        - user
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the user to update
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/UserAccount'
+                - type: object
+                - required: [username, email, color]
+        required: true
+      responses:
+        '200':
+          description: User updated
+        '400':
+          description: Invalid request body
+        '403':
+          description: Insufficent permissions
+        '500':
+          description: Internal server error
+  /session:
+    get:
+      summary: Determine if there is an active user session
+      tags:
+        - session
+      responses:
+        '200':
+          description: User is logged in
+        '404':
+          description: User is not logged
+    post:
+      summary: Start a session for the given user
+      tags:
+        - session
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/UserAccount'
+                - type: object
+                  required: [username, password]
+        required: true
+      responses:
+        '201':
+          description: Session started
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreationResponse'
+              example:
+                message: Session started
+                content: /api/session/{sessionToken}
+        '400':
+          description: Invalid request body
+        '500':
+          description: Internal server error
+    delete:
+      summary: Delete the current user's session
+      tags:
+        - session
+      responses:
+        '200':
+          description: Session deleted
+        '404':
+          description: No user session to delete
+        '500':
+          description: Failed to delete session
+  /list:
+    post:
+      summary: Create a new list for the logged in user
+      tags:
+        - list
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/List'
+                - type: object
+              required: [name, color]
+        required: true
+      responses:
+        '201':
+          description: List Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreationResponse'
+              example:
+                message: List Created
+                content: /list/{listId}
+        '400':
+          description: Invalid request body
+        '401':
+          description: No logged in user
+        '500':
+          description: Internal server error
+  /list/{id}:
+    patch:
+      summary: Update the list's metadata (e.g name, color, whether time tracking is enabled, etc.)
+      tags:
+        - list
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the list to update
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/List'
+                - type: object
+        required: true
+      responses:
+        '200':
+          description: List updated
+        '400':
+          description: Invalid request body
+        '401':
+          description: No logged in user
+        '404':
+          description: List not found
+        '500':
+          description: Internal server error
+    delete:
+      summary: Delete the list
+      tags:
+        - list
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the list to delete
+          schema:
+            type: string
+      responses:
+        '200':
+          description: List deleted
+        '400':
+          description: List not found
+        '401':
+          description: No logged in user
+        '500':
+          description: Internal server error
+  /list/{id}/tag:
+    post:
+      summary: Add a new tag to the list
+      tags:
+        - tags
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the list to add the tag to
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/Tag'
+                - type: object
+                  required: [name, color]
+        required: true
+      responses:
+        '201':
+          description: Tag Created
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CreationResponse'
+              example:
+                message: Tag created
+                content: /item/{id}/tag/${tagId}
+        '400':
+          description: Invalid request data
+        '401':
+          description: No logged in user
+        '500':
+          description: Internal server error
+  /list/{id}/tag/{tagid}:
+    patch:
+      summary: Update the tag
+      tags:
+        - tags
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the list that contains the tag
+          schema:
+            type: string
+        - name: tagid
+          in: path
+          required: true
+          description: The ID of the tag to update
+          schema:
+            type: string
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/Tag'
+                - type: object
+        required: true
+      responses:
+        '200':
+          description: Tag updated
+        '400':
+          description: Invalid request data
+        '401':
+          description: No logged in user
+        '500':
+          description: Internal server error
+    delete:
+      summary: Delete the tag
+      tags:
+        - tags
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The ID of the list to delete the tag from
+          schema:
+            type: string
+        - name: tagid
+          in: path
+          required: true
+          description: The ID of the tag to delete
+          schema:
+            type: string
+      responses:
+        '200':
+          description: Tag Deleted
+        '400':
+          description: Invalid request data
+        '401':
+          description: No logged in user
+        '500':
+          description: Internal server error
+  /list/{id}/section:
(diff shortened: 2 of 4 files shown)


app/globals.css
```diff
@@ -1,3 +1,6 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+html {
+  @apply transition-colors duration-500 ease-in-out;
+}
```
radiantBear: I don't think we should add this transition; it results in some strange looking behavior when changing the theme on other pages (see https://github.com/user-attachments/assets/a87a6bcc-aaa6-4586-8ece-f63353c36a33)

app/providers.tsx
```diff
@@ -21,6 +21,7 @@
 import { NextUIProvider } from '@nextui-org/react';
 import { ThemeProvider } from 'next-themes';
 import { useRouter } from 'next/navigation';
+import React from 'react';
```
radiantBear: I thought the last few versions of React implicitly import this object in all files and it works just fine for me without this import. Was this missing import throwing errors for you or can it be removed again?

components/ThemeSwitcher.tsx
```diff
@@ -16,6 +16,8 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+'use client';
+
```
radiantBear: Why is the `'use client';` directive needed now?

app/globals.css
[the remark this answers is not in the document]
sabri414: Yeah, I understand. The reason behind what I did is that it ensures backgrounds, text colors, and images fade nicely when switching themes (no harsh flicker).

public/screenshots/Filtring Dark.png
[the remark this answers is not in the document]
sabri414: Oh Yeah! I totally forgot that. It's coding 101, sorry, should've known better

app/providers.tsx
sabri414: I added it without it causing an error on my end. I think we can remove it

app/page.tsx
```diff
@@ -16,6 +16,56 @@
...
+      'Filter tasks by tag, priority, or due date to manage your workload efficiently.',
+    imageBaseName: 'filtering'
+  },
+  {
+    title: 'Manual Sorting',
+    description:
+      'Reorder tasks manually by dragging and dropping them within each section.',
+    imageBaseName: 'sorting'
+  }
+];
```
radiantBear: Is leaving this as a variable the best option? Or should it be inlined to several `<FeatureBlock />` components below or extracted to a JSON file?

Cc @lilyvesstible & @A-man37 for thoughts - let's use this case to establish a rule of thumb for future static pages

public/screenshots/timeTracking.light.png
radiantBear: Some of these pictures aren't high enough resolution - let's upload some better ones

app/page.tsx
```diff
@@ -16,6 +16,60 @@
...
+        </h1>
+        <p className='mt-3 text-lg'>
+          Stay organized with time tracking, tagging, and smart filtering.
+        </p>
+      </header>
+
+      {FEATURES.map((f, i) => (
+        <FeatureBlockIndexProvider key={f.title} index={i}>
+          <FeatureBlock {...f} />
+        </FeatureBlockIndexProvider>
```
radiantBear: The `index` prop to `<FeatureBlockIndexProvider>` can also be removed: https://react.dev/learn/passing-data-deeply-with-context#using-and-providing-context-from-the-same-component

app/page.tsx
```diff
@@ -16,6 +16,60 @@
...
 
+import FeatureBlock, {
+  FeatureBlockIndexProvider
+} from '@/components/FeatureBlock';
+
+const FEATURES = [
+  {
+    title: 'Time Tracking',
+    description:
+      'Track expected vs elapsed time for every task with quick Start and Pause buttons.',
```
radiantBear: Small nitpick; I think this would flow better without the word "quick" and make Start & Pause lowercase - what do you think?

app/page.tsx
```diff
@@ -16,6 +16,60 @@
...
+  {
+    title: 'Tag System',
+    description:
+      'Organize tasks with customizable colored tags for subjects, projects, or contexts.',
+    imageBaseName: 'tagSystem'
+  },
+  {
+    title: 'Priority System',
+    description:
+      'Mark tasks as Low, Medium, or High priority to stay focused on what matters most.',
```
radiantBear: Same small style comment here - I think normal sentence case (without capitalized Low, Medium, High) would be ideal here

app/page.tsx
```diff
@@ -16,6 +16,60 @@
...
+  {
+    title: 'Filtering Feature',
+    description:
+      'Filter tasks by tag, priority, or due date to manage your workload efficiently.',
+    imageBaseName: 'filtering'
+  },
+  {
+    title: 'Manual Sorting',
+    description:
+      'Reorder tasks manually by dragging and dropping them within each section.',
```
radiantBear: Since we don't explicitly discuss auto-sorting anywhere else, I think it might make sense to include that here - thoughts?
> Tasks are intelligently sorted based on due date and priority. If you need extra control, however, you're also free to arrange them manually.

radiantBear (commented):
Thank you Sabri, this is looking good! Thanks for the work on starting to incorporate `Context` in particular; I left some more details about that in my review. I also added a few comments about the specific wording for each description, but the rest looks pretty solid!

app/page.tsx
sabri414: Ahh, I see, just fixed that

components/FeatureBlock.tsx
```diff
@@ -0,0 +1,99 @@
...
+type FeatureBlockProps = {
+  title: string;
+  description: string;
+  imageBaseName: string;
+};
+
+// Context to provide positional info (e.g., index) without prop drilling
+export const FeatureBlockIndexContext = createContext<number>(0);
+
+export function FeatureBlockIndexProvider({
```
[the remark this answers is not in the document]
radiantBear: Sounds good; thanks Sabri! Can the `FeatureBlockIndexProvider` function be deleted now then?

radiantBear (commented):
Thanks again for putting this together Sabri! This is looking pretty good; I left a few last comments, but overall I'm pretty comfortable with how things're looking - things are well designed and clear.

Since you're creating a new component, I think this would be a really great time to figure out component testing with Jest! We can talk about that more tomorrow

app/page.tsx
A-man37: I think that it makes sense for us to include any kind of variable data in JSON files. This way the actual code functionality is unaffected if someone chooses update the content of the homepage. I think that this would be a good rule of thumb going forward for any variable content that we plan to display to the user.

Ryangman (changes requested):

components/FeatureBlock.tsx
```diff
@@ -0,0 +1,87 @@
...
+}: FeatureBlockProps) {
+  const index = useContext(FeatureBlockIndexContext);
+
+  return (
+    <section
+      className={`flex flex-col items-center gap-8 md:gap-16
+                  md:flex-row ${align === 'flipped' ? 'md:flex-row-reverse' : ''}`}
+    >
+      <div className='w-full md:w-1/2'>
+        <div className='rounded-xl border shadow-md overflow-hidden border-gray-200 dark:border-gray-800'>
```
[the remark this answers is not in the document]
A-man37: I would recommend using HeroUI. From a brief look at the documentation, it looks like you would only need to write the image object once instead of having a block of code for both lightmode and darkmode images. This should help prevent errors in future code maintenance.

The Featureblock object looks great on the homepage.

package.json
```diff
@@ -54,12 +58,12 @@
     "eslint-plugin-react": "^7.37.2",
     "eslint-plugin-react-hooks": "^4.6.2",
     "eslint-plugin-unused-imports": "4.1.4",
+    "identity-obj-proxy": "^3.0.0",
```
radiantBear: What is this used for?

components/__mocks__/next/image.js
radiantBear: Is this file being used? I'm not familiar with how Jest handles mocks like this, but it looks like `next/image` isn't being imported in `FeatureBlock.tsx` anymore, so I assume this wouldn't do anything? If so, it should be removed

components/__tests__/FeatureBlock.test.js
```diff
@@ -0,0 +1,66 @@
...
+        align='flipped'
+      />
+    );
+
+    const section = container.querySelector('section');
+
+    expect(section.className).toMatch(/md:flex-row-reverse/);
+  });
+
+  it('renders both light and dark images', () => {
```
radiantBear: Instead of just testing that both images are there, I think it would be better to manipulate the site theme and test that only the correct image is visible (probably 2 separate tests)

components/FeatureBlock.tsx
```diff
@@ -18,7 +18,7 @@
 
 'use client';
 
-import Image from 'next/image';
+import { Image as HeroImage } from '@nextui-org/image';
```
Ryangman: Since we are no longer using nextui, this will need to use the version from heroui. It might make sense to rebase your changes on master to ensure it looks as expected with heroui components/styling before this is merged

jest.setup.js
```diff
@@ -0,0 +1,25 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import '@testing-library/jest-dom';
+
+global.Response = class {
```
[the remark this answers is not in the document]
sabri414: global.Response isn’t used by the new FeatureBlock tests themselves. It was added because once Jest started running with proper ESM/Babel transforms, the backend test suite began throwing ReferenceError: Response is not defined.

This wasn’t caused by the homepage changes; it’s because the backend tests depend on the Web Fetch API (Response.json, Headers, etc.), which doesn’t exist in Node’s Jest environment unless explicitly mocked.

Since enabling Jest fully exposed those failures, I added the global mock to allow the rest of the test suite to run locally.

That said, I agree this shouldn’t live in this PR if we’re keeping the testing infrastructure untouched for now. If we move the Jest configuration work (including the global.Response mock) into its own dedicated PR; where we also fix the backend tests, I can remove it from this PR so the homepage remains isolated.

Happy to adjust based on how we want to structure the follow-up work!

components/FeatureBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const themedSrc =
+    theme === 'dark'
+      ? `/screenshots/${imageBaseName}.dark.png`
+      : `/screenshots/${imageBaseName}.light.png`;
+
+  return (
+    /**
+     * Controls whether the feature block layout is left-to-right or reversed.
+     * When `align` is "flipped", the screenshot appears on the right instead of the left.
+     */
```
radiantBear: Thanks for adding the block comments! This comment doesn't seem accurate though; the following JSX is the entire component, not just logic for whether to display the block reversed or not. I believe it should be omitted

components/FeatureBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+  description,
+  imageBaseName,
+  align = 'default'
+}: FeatureBlockProps) {
+  const { theme } = useTheme();
+  /**
+   * Determines which screenshot to display based on the current theme.
+   *
+   * @returns The full URL of either the light-mode or dark-mode screenshot.
+   */
```
radiantBear: Thanks for adding the block comments! This one is more verbose than the code itself, though, so I think it should be omitted

components/FeatureBlock.tsx
```diff
@@ -0,0 +1,102 @@
...
+ * Displays a title, description, and a themed screenshot for a given feature.
+ * The image automatically switches between light and dark variants based
+ * on the active theme from `next-themes`. Layout can be optionally flipped
+ * for alternating left/right visual alignment.
+ *
+ * @param {string} title - The name of the feature being described.
+ * @param {string} description - A short explanation of what the feature does.
+ * @param {string} imageBaseName - Base filename used to load the appropriate
+ * light or dark screenshot (e.g., "timeTracking" → "timeTracking.light.png").
+ * @param {"default" | "flipped"} [align="default"] - Determines whether the
```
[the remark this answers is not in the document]
radiantBear: In addition to saying
> Determines whether the image appears on the left or the right of the text.

It would be helpful to include which one is the default, e.g.
> Determines whether the image appears on the left (default) or the right (flipped) side of the text.

app/page.tsx
```diff
@@ -14,8 +14,49 @@
...
+ * `/public/data/features.json` to keep the page content configurable without modifying
+ * application code.
+ *
+ * Each feature is rendered using the `FeatureBlock` component, which handles
+ * themed screenshots (light/dark mode) and optional layout flipping. The layout
+ * alternates between left- and right-aligned images to create a visually
+ * balanced presentation.
+ *
+ * @returns The homepage layout containing the hero section and a list of
+ * feature showcase blocks.
```
radiantBear: This part of the comment looks like it's for the `Page` function. Please move it to line 39, right above
```ts
export default function Page() {
```

components/FeatureBlock.tsx
```diff
@@ -37,19 +36,15 @@ type FeatureBlockProps = Readonly<{
   align?: 'default' | 'flipped';
 }>;
 /**
- * Renders a feature showcase block on the homepage.
+ * Displays a homepage feature section with a screenshot, title, and description.
```
radiantBear: This component is well-designed for general use anywhere a feature highlight may be used and is appropriately placed in the main `components/` directory for use throughout Tasktix.

Including "homepage" in this comment implies tighter coupling to the homepage than this component is designed for. This documentation should be more general to apply anywhere this component may be used. 

Also, while the type signature of each parameter doesn't need to be documented in the comment, please do still describe each parameter and how it should be used.

This might look something like this:

```tsx
/**
 * Highlights a key feature with a image showing the feature in action and a description explaining the feature
 * 
 * @param title - The name of the feature being highlighted
 * @param description - A short explanation of the feature's capabilities & value
 * @param imageBaseName - The base filename used to load the appropriate light- or dark-theme screenshot (e.g. `imageBaseName="timeTracking"` will load the image `"public/screenshots/timeTracking.light.png"` in light mode and `"public/screenshots/timeTracking.dark.png"` in dark mode)
 * @param align - Determines whether the image appears on the left (`default`) or the right (`flipped`) side of the text
 */
export default function FeatureBlock({
    ...
```

radiantBear (commented):
Looking good, thanks Sabri! Don't forget to merge the changes from `main` back into your branch to ensure everything's up-to-date. I tried leaving a few more-specific suggestions this time; let me know what you think

components/FeatureBlock.tsx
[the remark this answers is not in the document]
sabri414: fixed

components/FeatureBlock.tsx
radiantBear: Thanks Sabri; nice improvements here! Especially appreciate the `align` prop's description "... appears before (`default`) or after (`flipped`)" - very nice wording that accurately captures both desktop & mobile displays!

radiantBear (approved):
Thank you Sabri; this looks great! Very excited to have a homepage that shows off our key differentiators!

A-man37 (approved):
Thanks for getting this updated Sabri. The HeroUI implementation looks great and I really like the testing that you are doing. I think that this PR is ready to merge! Great work.

radiantBear (approved):

PR #16, opened 2025-11-09
chore: migrate deprecated nextui dependencies

### LINKED ISSUE #4: Update nextui dependencies to heroui

Ryangman:
Migrates NextUI to HeroUI, including peer dependencies (Tailwind, Postcss). Fixed all styling issues introduced by the migration, and several preexisting bugs. Updated Dockerfile to support architecture dependent packages. Minimal changes to user interface. 

## Related Issues

Closes #4 

## Changes
- Migrate NextUI->HeroUI
- update to Tailwind v4
- Update postcss/tailwind configs and `globals.css`

### Fixed Several Styling Issues created by Migration
- Username/password input changed to 'bordered'
- Replaced uses of `w-100` with `w-full`
- Fixed responsiveness of "Add Task button"
- Changed Tag item from `Card` to `Button` to support Popovers (HeroUI only allows popovers on interactive items)
- Removed no longer valid `key` prop of  `SelectItem`

## Reviewer Notes

In `app/providers.tsx`, I had to defer the import of HeroUIProvider to the client, due to the errors noted in [this issue](https://github.com/heroui-inc/heroui/issues/5612). The solution they found of reverting NextJS versions did not work for us. Will be something to look into in future, if we consider updating to a modern NextJS version.
```ts  
const HeroUIProvider = dynamic(
    () => import('@heroui/react').then(mod => mod.HeroUIProvider),
    { ssr: false }
  );
```

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -1,4 +1,6 @@
 @import "tailwindcss";
 @config "../tailwind.config.ts";
+
+/* Apply safelist for programatically used classes see https://tailwindcss.com/docs/detecting-classes-in-source-files */
 @source inline('!{,}{bg,text}-{danger,warning,success}');
 @source inline('!{,}{bg,text}-{pink,red,orange,amber,yellow,lime,green,emerald,cyan,blue,violet}-500');
--- a/app/body.tsx
+++ b/app/body.tsx
@@ -58,7 +58,10 @@ interface BodyProps {
   isLoggedInAtStart: boolean;
 }
 
-export default function Body({ children, isLoggedInAtStart }: BodyProps) {
+export default function Body({
+  children,
+  isLoggedInAtStart
+}: Readonly<BodyProps>) {
   const [isLoggedIn, setIsLoggedIn] = useState(isLoggedInAtStart);
 
   _setLoggedIn = () => setIsLoggedIn(true);
--- a/components/SearchBar/InputElement.tsx
+++ b/components/SearchBar/InputElement.tsx
@@ -83,15 +83,15 @@ export default function InputElement({
         >
           {inputOption.type === 'Select'
             ? inputOption.selectOptions?.map(option => (
-              <SelectItem
-                key={option.name}
-                className={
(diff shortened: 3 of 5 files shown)


app/globals.css
```diff
@@ -1,3 +1,4 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import "tailwindcss";
+@config "../tailwind.config.ts";
+@source inline('!{,}{bg,text}-{danger,warning,success}');
+@source inline('!{,}{bg,text}-{pink,red,orange,amber,yellow,lime,green,emerald,cyan,blue,violet}-500');
```
radiantBear: Could you add a comment explaining why these are inlined?

radiantBear (dismissed):
Thanks Ryan, it's great to have this up-to-date!

I discovered that the `<Input>` component's invisible underline in dark mode is also an issue for the UI components for creating a new tag. We may want to consider finding a way to override HeroUI's underline style in general so we can return to the underlined `<Input>`s in the login UI as well.

I'll leave that up to you, though - I don't think it's a particularly major issue to fix in this PR

app/globals.css
Ryangman: added a link to the Tailwind documentation, long story short, that is the preferred and supported way to implement a safelist in v4.

radiantBear: Thanks! I actually just meant adding a comment about what we discussed about why we have them safelisted (i.e. they're dynamically generated and Tailwind's tree shaking will remove them and leave things unstyled unless they're safelisted).

Interesting that this is a problem at all; I just read that Tailwind page and it makes it sound like they should've been included. Maybe the issue is that the function is imported from another file?

radiantBear (approved):
Looking good!

A-man37 (approved):
This looks really great, Ryan. I am not totally familiar with implementing heroui, but this code looks correct. I can see that you have updated the file headers appropriately and have simplified some of the javascript.

PR #17, opened 2025-11-10
feat(ui): add user profile page

lilyvesstible:
Adding profiles, and a profile preferences page where the user can change details about their profile.

## Related Issues
Closes Tasktix #11

## Changes

- [X] Added profile preferences page
- [X] Updated sign out button, to now be an avatar dropdown menu.

### Tests added

- Testing setUsername()
- Testing setEmail()

## Reviewer Notes
Further implementation is needed for adding profile pictures, and changing other user settings.

### DIFF
--- a/components/ConfirmedTextInput.tsx
+++ b/components/ConfirmedTextInput.tsx
@@ -59,10 +59,10 @@ export default function ConfirmedTextInput({
   variant?: InputProps['variant'];
   updateValue: (value: string) => unknown;
 }) {
-  const [newName, setNewName] = useState(value);
-  const [prevName, setPrevName] = useState(value);
+  const [newValue, setNewValue] = useState(value);
+  const [prevValue, setPrevValue] = useState(value);
 
-  useEffect(() => setPrevName(value), [value]);
+  useEffect(() => setPrevValue(value), [value]);
 
   return (
     <form
@@ -70,7 +70,7 @@ export default function ConfirmedTextInput({
       data-testid={`confirmed-input-${label ?? 'value'}`}
       onSubmit={e => {
         e.preventDefault();
-        updateValue(newName);
+        updateValue(newValue);
       }}
     >
       <Input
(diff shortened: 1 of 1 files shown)


components/ConfirmedTextInput.tsx
```diff
@@ -16,25 +16,45 @@
...
+
+import { Button, Input, InputProps } from '@heroui/react';
 import { useEffect, useState } from 'react';
 import { Check } from 'react-bootstrap-icons';
 
-export default function Name({
+/**
+ * A text input that automatically fills with text. When changed, a confirmation button appears. The confirmation button executes a function. Used for changing information in the database.
+ *
+ * @param name The text that should automatically fill the field. Will also be used as an input for the function.
```
radiantBear: Thanks for generalizing this component so it's not just for names! Could you please change the `name` and `updateName` props to `value` and `updateValue` to align with this generalization?

app/body.tsx
```diff
@@ -152,8 +157,18 @@ function AccountButton({ isLoggedIn }: { isLoggedIn: boolean }) {
...
 
   return (
-    <Button key='signOut' color='primary' variant='flat' onPress={handleClick}>
-      Sign Out
-    </Button>
+    <Dropdown>
+      <DropdownTrigger>
+        <Avatar key='profile' color='primary' />
+      </DropdownTrigger>
+      <DropdownMenu aria-label='Static Actions' color='primary'>
```
radiantBear: I think it would be cool to use `user.color` instead of just `primary` for this button. I think that'll require some further refactoring to how the logged-in user is stored in React state, though, so we can wait on that.

app/profile/components/SetUserProps.tsx
```diff
@@ -0,0 +1,107 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+'use client';
+
+import { useState } from 'react';
+import { addToast } from '@heroui/react';
+
+import { default as api } from '@/lib/api';
```
radiantBear: Like SonarQube points out here, you can just `import api from...` instead of `import { default as api } ...`

app/profile/components/SetUserProps.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+import SetUserProps from './SetUserProps';
+
+jest.mock('@/lib/api');
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
+describe('Set user props functions', () => {
```
radiantBear: Our convention has been to use `describe` blocks to separate tests for each function. Instead of `'Set user props functions'`, could you please change this to match the function name (`SetUserProps`)?

app/body.tsx
lilyvesstible: I think we should wait to implement this until we implement the UI to change the user's color. I put a todo comment in there though!

A-man37 (dismissed):
Fantastic work. I am excited to see this page implemented in the project. I don't really have any feedback to give at this time. This is going to be a great base for user profiles that we can expand on in the future if we want to.

A-man37 (approved):
Looks good!

PR #19, opened 2025-11-14
refactor!: use Prisma ORM instead of raw SQL

### LINKED ISSUE #12: Add ORM for Database Actions

radiantBear:
This PR migrates from using raw SQL queries to modify the database to a modern TypeScript-first ORM. This will help decrease bugs caused by unexpected edge cases in object-to-relation (and vice versa) mapping, improve type safety for database interactions, and decrease maintenance costs by removing an entire component of the backend. It also makes database migrations easier as we prepare for production deployment.

## Related Issues
- Closes #12 

## Changes

- [x] Added an image to the test infra to execute Prisma CLI commands against the database (e.g. migrations, seeding, etc.) since this isn't included in the production `web` image
- [x] Added Danger rules to ensure the Prisma schema and migrations stay in sync
- [x] Converted database seeding to TypeScript + Prisma instead of raw SQL
  - Database function abstraction preserved (instead of inlining Prisma queries) to minimize maintenance costs
- [x] Fixed database connections to maintain 1 persistent connection instead of creating a new connection for every query
- [x] Removed models representing raw database rows
- [x] Removed functions converting raw database rows to TypeScript objects
- [x] Removed raw SQL execution library

### Extras
- [x] Fixed some database functions' type signatures to more accurately reflect their needs/results
- [x] Updated `package.json` to reflect most of the project uses ES modules, not CommonJS

## Reviewer Notes

> [!IMPORTANT]
> This PR includes **breaking database changes**. You will need to run `npm run teardown -- -v` before starting the server for the first time

Does the updated "getting started" documentation adequately explain how to set up a new system with a database?


lib/database/listItem.ts
```diff
@@ -18,154 +18,100 @@
...
-  if (!result) return false;
-
-  return extractListItemsFromRows(result);
+export async function getListItemsByUser(userId: string): Promise<ListItem[]> {
+  const result = await prisma.item.findMany({
+    where: { section: { list: { members: { some: { userId } } } } },
+    include: { tags: true, assignees: { include: { user: true } } }
+  });
+
+  return result;
```
Ryangman: Does this query no longer need error handling to return false?

lib/database/list.ts
```diff
@@ -184,161 +135,126 @@ export async function getIsListAssignee(
...
 }
 
 export async function getTagById(id: string): Promise<Tag | false> {
-  const sql = `
-    SELECT * FROM \`tags\`
-    WHERE \`tags\`.\`t_id\` = :id;
-  `;
-
-  const result = await query<DB_Tag>(sql, { id });
+  const result = await prisma.tag.findFirst({
```
Ryangman: this and `getListBySectionId` seem like situations that would warrant `findUnique` rather than `findFirst`

Ryangman (approved):
Thanks for all the effort in this! it looks like a much simpler database interaction pattern. Changes look good to me, I mentioned a couple areas that seemed inconsistent/in need of documentation and some potential future improvements.

app/api/list/[id]/tag/[tagId]/route.ts
```diff
@@ -42,6 +42,7 @@ export async function PATCH(
 
   const tag = await getTagById(params.tagId);
 
+  // TODO: need to handle security gap: editing someone else's tag
```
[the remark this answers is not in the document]
Comment (radiantBear) [reply app/api/list/[id]/tag/[tagId]/route.ts]: Good thought! SonarQube does flag them, but maybe we should add a check to Danger, too, to make it more visible. What do you think?

lib/database/listItem.ts
radiantBear: I believe the error handling was just necessary for TypeScript because of `query`'s type signature. If there are no matching results, `findMany` [should just return an empty list](https://www.prisma.io/docs/orm/reference/prisma-client-reference#return-type-2).

lib/database/list.ts
radiantBear: Good catch! Unfortunately Prisma's typing doesn't support `findUnique` for unique fields on the M side of 1:M relationships (e.g. in `getListBySectionId`), but it does work here!

package.json
```diff
@@ -11,21 +12,28 @@
     "danger": "danger ci",
     "local:build": "next build",
     "local:dev": "next dev",
+    "prisma:introspect": "prisma db pull",
```
Ryangman: It would be useful to document the process of database changes. (i.e what each of these commands do and how to use them properly update the schema/tests for any database changes).

Ryangman (approved):
Changes look good to me now

Ryangman (approved):
documentation looks good. This seems ready to mainline

prisma/migrations/20251109044725_init/migration.sql
```diff
@@ -0,0 +1,146 @@
...
+
+    INDEX `Session_userId_idx`(`userId`),
+    PRIMARY KEY (`id`)
+) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+-- CreateTable
+CREATE TABLE `List` (
+    `id` CHAR(16) NOT NULL,
+    `name` VARCHAR(64) NOT NULL,
+    `color` ENUM('Pink', 'Red', 'Orange', 'Amber', 'Yellow', 'Lime', 'Green', 'Emerald', 'Cyan', 'Blue', 'Violet') NOT NULL,
```
A-man37: Are we allowing lists to be a separate color from the User's color? If I understand correctly, that is what you have here. I think that makes sense, because a user can assign different colors to different tasks as a way of specifying what type of test it is. Even better, somewhere along the line we could add functionality for the system to assign colors to tasks automatically. Like making a task color purple to represent multiple collaborators, or orange to represent a recurring task.

PR #27, opened 2025-11-29
feat: enable adding members to collaborate on lists

### LINKED ISSUE #23: Allow adding other users to lists
Users should be able to add others to their list to collaborate on tasks. The initial version will simply allow any other user to be added by username

radiantBear:
This PR allows users to add others to their list so Tasktix can be used to manage tasks for group/team projects, not just individual work. Once a list has more than 1 member in it, the feature to assign particular members to specific tasks (already implemented) is enabled.

## Related Issues
- Closes #23
- Related to #13 

## Changes
- [x] Added a tab to the settings page to add members & update member permissions
- [x] Added backend support for adding members & changing their permissions
- [x] Added permissions checks to validate actions like creating list items, completing them, and adding assignees
- [x] Refactored settings component to extract implementation of each tab/section of settings
- [x] Fixed the "assign user" popover from the HeroUI migration (popover target must be a button now)

## Testing Coverage
No special needs for testing this PR. Only unit tests were added, so `npm run test:jest` (i.e. skipping Cypress tests) is sufficient to run all tests.

### Tests added
- Unit tests for the new `<MemberSettings />` component
- Unit tests for the new API routes

### Tests broken
N/A

### Tests needed
- Tests for the other refactored `<___Settings />` components
    - Not included in this PR because no logic was changed for them

## Reviewer Notes
We should discuss the permissions system to design an alternative that's more robust and simple to use. Not included in this PR because this PR only uses the current permissions system; it didn't create it.

Let me know if you'd like to see unit tests for the other `<___Settings />` components added.

### DIFF
--- a/app/api/list/[id]/member/[userId]/route.ts
+++ b/app/api/list/[id]/member/[userId]/route.ts
@@ -32,10 +32,10 @@ const PatchBody = ZodListMember.omit({ listId: true, userId: true }).partial();
  *
  * @param params.id The list to modify the member for
  * @param params.userId The user to modify membership data for
- * @param request.canAdd Whether the user can add items to the list
- * @param request.canRemove Whether the user can delete items from the list
- * @param request.canComplete Whether the user can mark items as completed
- * @param request.canAssign Whether the user can assign items to other users
+ * @param request.canAdd [optional] Whether the user can add items to the list
+ * @param request.canRemove [optional] Whether the user can delete items from the list
+ * @param request.canComplete [optional] Whether the user can mark items as completed
+ * @param request.canAssign [optional] Whether the user can assign items to other users
  */
 export async function PATCH(
   request: Request,


app/api/list/[id]/member/[userId]/route.ts
```diff
@@ -0,0 +1,76 @@
...
+import { getUser } from '@/lib/session';
+
+const PatchBody = ZodListMember.omit({ listId: true, userId: true }).partial();
+
+/**
+ * Update a member's permissions for a specific list
+ *
+ * @param params.id The list to modify the member for
+ * @param params.userId The user to modify membership data for
+ * @param request.canAdd Whether the user can add items to the list
```
Comment (lilyvesstible) [inline app/api/list/[id]/member/[userId]/route.ts]: This might be a personal ignorance, but I don't see where any props for request are defined. I do see that they are used in line 63 onwards, and this may be the definition. If these props are defined elsewhere it should be made clear, because in that case the documentation might be misleading. If this is the correct documentation, ignore this.

app/api/list/[id]/member/route.ts
```diff
@@ -0,0 +1,71 @@
...
+
+const PostBody = ZodUser.pick({ username: true });
+
+/**
+ * Add a user as a member of the list so they can view it and be assigned items to
+ * complete. When first added, members will have no permissions & must be granted them
+ * after
+ *
+ * @param params.id The list to add the member to
+ * @param request.username The username of the member to add to the list
```
Comment (lilyvesstible) [inline app/api/list/[id]/member/route.ts]: Same as previous comment.

app/api/list/[id]/member/[userId]/route.ts
Comment (radiantBear) [reply app/api/list/[id]/member/[userId]/route.ts]: Thanks for calling that out Lily! You're absolutely right that the documentation should be clear enough to follow without substantial prior knowledge.

`request` is one of the variables passed to `PATCH()` (on line 41). `request` is the JavaScript `Request` object Next.JS generates from the HTTP request and passes to the `PATCH()` function. This JS `Request` object includes a JSON string with the request body that the user sent. This request body should include `canAdd`, `canRemove`, etc. fields. I tried to split up the request body's expected fields so I could document the purpose of each field, represented via the `request.canAdd`, etc. parameters here. Do you think there is a better way I could convey that meaning in the documentation?

Ryangman (dismissed):
These changes look good! Some notable holes for future development are improving styling for the MemberSettings Page. Adding a DELETE route to remove a member from a list, and deciding on proper access control for list members.  This is ready to be merged.

sabri414 (approved):
Looks good!

PR #31, opened 2025-12-03
chore: update DeepSource config

radiantBear:
This PR improves DeepSource's configuration so that we can see code coverage at a glance & so that the config accurately reflects the repo architecture.

## Changes

- [x] Added DeepSource test coverage tracking
- [x] Updated test patterns so DeepSource accurately identifies all test files
- [x] Updated DeepSource Docker scanning to include the test Dockerfile

## Testing Coverage

To test whether this PR's changes are successful, check the DeepSource comment in this PR or the DeepSource dashboard for this PR to see if coverage is tracked.


Ryangman (approved):
These changes look good. Something that may make sense to throw in this PR would be setting up sonarqube's `//TODO:` flagging with Danger. But that can also be saved for later.

radiantBear:
> setting up sonarqube's `//TODO:` flagging with Danger

I'm not sure I'm following - do you not think SonarQube's flagging like this in the "Files changed" tab isn't sufficient? If not, what're exactly should we add?
image

My best guess is that you're hoping it'll auto-reject PRs with TODOs in them, but I think allowing human review like this is good - sometimes noting a TODO is good to quickly find things later, but the change shouldn't be part of the same PR.

radiantBear:
Good point, Ryan. SonarQube does only flag TODOs in files that were edited in the current PR. I just created #35 to address this in another PR

PR #45, opened 2026-01-12
fix(ui): route authenticated users to /list when clicking logo

sabri414:
Improves navigation UX by routing logged-in users to their task list when clicking the Tasktix logo, preventing users from getting stuck on the marketing homepage after login.

### DIFF
--- a/components/Filter/FilterText.tsx
+++ b/components/Filter/FilterText.tsx
@@ -1,0 +1,157 @@
+/**
+ * Tasktix: A powerful and flexible task-tracking tool for all.
+ * Copyright (C) 2025 Nate Baird & other Tasktix contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+import { formatTime } from '@/lib/date';
+
+import {
+  DateFilterOperator,
+  FilterInputState,
+  FilterState,
+  OptionFilterOperator
+} from './types';
+
+export const LABEL_COLOR = 'text-primary-600' as const;
+export const OPERATOR_COLOR = 'text-primary' as const;
+export const PRIMITIVE_COLOR = 'text-green-800 dark:text-green-200' as const;
+export const STRING_COLOR = 'text-warning-700 dark:text-warning-200' as const;
+
+/**
+ * Pretty-prints the filter state as syntax-highlighted query language
+ *
+ * @param filters The filters to pretty-print
+ */
+export default function FilterText({ filters }: { filters: FilterState }) {
+  return (
+    <span className='font-mono'>
+      {filters.filters.map((f, i) => (
+        <>
+          {f && 'filters' in f ? (
+            <>
+              {'( '}
+              <FilterText key={f.id} filters={f} />
+              {' )'}
+            </>
+          ) : (
+            <FilterInputText key={f.id} filter={f} />
+          )}
+          {i < filters.filters.length - 1 && (
+            <>
+              {' '}
+              <span className={OPERATOR_COLOR}>
+                {filters.operator.toUpperCase()}
+              </span>{' '}
+            </>
+          )}
+        </>
+      ))}
+    </span>
+  );
+}
+
+/**
+ * Pretty-prints the filter as syntax-highlighted query language
+ *
+ * @param filter The filter to pretty-print
+ */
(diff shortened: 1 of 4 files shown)


app/page.tsx
```diff
@@ -14,49 +14,18 @@
...
-          Tasktix — Smarter Task Tracking
-        </h1>
-        <p className='mt-3 text-lg'>
-          Stay organized with time tracking, tagging, and smart filtering.
-        </p>
-      </header>
+import { getUser } from '@/lib/session';
+
+export default async function Home() {
+  const user = await getUser();
```
radiantBear: I'm debating whether it matters because I don't think there's a way to end up here from Next.js's internal routing (which may break this functionality because of caching). Regardless, for consistency, I think it would be good to use `useAuth()` instead of `getUser()` here, just like app/body.tsx

radiantBear (commented):
Thanks Sabri, this is a good start! I left some feedback below

app/page.tsx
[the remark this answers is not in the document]
sabri414: I’ve added a file-level comment to about/page.tsx clarifying its role as the public marketing page at /about. This page is intentionally thin and mostly static, with behavior delegated to FeatureBlock (which already has test coverage). If the team would like a page-level render test added here as well, I’m happy to do so, just let me know the preferred testing pattern.

app/page.tsx
radiantBear: Oh, nice! Thanks for the reminder; this of course makes more sense given the client/server component split.

radiantBear (commented):
Thanks Sabri! Left a few more thoughts below. Also, please don't forget to update your PR title to match Danger's expectations!

sabri414:
> Thanks Sabri! Left a few more thoughts below. Also, please don't forget to update your PR title to match Danger's expectations!

Thanks for the clarification, that makes sense. You’re right that app/page.tsx contains redirect logic of its own. Would you mind explaining to me why Danger is unhappy with my PR title?
I’ve added an automated test covering both redirect paths (/list vs /about) to guard against regressions. I’ve also replaced the non-ASCII arrows with -> and moved the documentation comment directly above Home() as suggested. Appreciate the careful review!

app/body.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+    replace: jest.fn(),
+    push: jest.fn(),
+    prefetch: jest.fn()
+  })
+}));
+
+jest.mock('next/image', () => ({
+  __esModule: true,
+  default: ({ alt, ...rest }: { alt?: string; src?: string }) => (
+    <span aria-label={alt} data-src={String(rest.src ?? '')} role='img' />
```
radiantBear: Defining a `...rest` parameter seems unnecessary since only `rest.src` is being used. Please simplify it to just `{ alt, src }`.

app/body.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+  DropdownMenu: ({ children }: React.PropsWithChildren) => (
+    <div>{children}</div>
+  ),
+
+  DropdownItem: ({ children, ...props }: Props) => (
+    <div {...props}>{children}</div>
+  ),
+
+  Avatar: () => <div />
+}));
```
radiantBear: This seems like pretty excessive mocking. Why is all of this needed? It looks like `useAuth` is the only mock that has a functional impact on testability. Where possible, default behavior should be used instead of mocks.

Note: there's a weird error with some HeroUI components when they're clicked on. This can be fixed by wrapping the SUT in:
```tsx
  /* YOUR SUT HERE */
</HeroUIProvider>
```

app/body.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+describe('Body', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
```
radiantBear: Since tests are changing mock implementations, we should use `jest.resetAllMocks()` here instead to ensure test isolation.

jest.config.cjs
```diff
@@ -157,7 +158,6 @@ const config = {
   // setupFiles: [],
 
   // A list of paths to modules that run some code to configure or set up the testing framework before each test
-  setupFilesAfterEnv: ['<rootDir>/jest.setup.ts'],
```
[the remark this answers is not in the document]
sabri414: When i add it Javascript analyzer complains

app/body.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+ */
+
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+
+import Body from '@/app/body';
+import { useAuth } from '@/components/AuthProvider';
+
+import '@testing-library/jest-dom';
+import 'whatwg-fetch';
```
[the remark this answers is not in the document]
sabri414: Yeah i just checked i don't need it anymore

app/body.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  useAuth: jest.fn()
+}));
+
+jest.mock('next/navigation', () => ({
+  useRouter: () => ({
+    replace: jest.fn(),
+    push: jest.fn(),
+    prefetch: jest.fn()
+  })
+}));
```
radiantBear: Why is this being mocked? Router functions shouldn't be called during the course of these tests anyway

sabri414: even though router methods aren’t invoked in the test, AccountButton calls useRouter() during render. In Jest this throws invariant expected app router to be mounted, so the mock prevents the render from failing.

PR #46, opened 2026-01-12
test: remove codacy from static checks

radiantBear:
Codacy is extremely noisy and hasn't contributed any significant issue identification/correction because it mostly relies on other tools that are in use, such as ESLint and Semgrep, that have more sane defaults than what Codacy picked. Using it increases developer burden without producing results, so it is being removed. The only impact to the repo contents is removing its badge from the README.


sabri414 (approved):
sweet, we will wave codacy goodbye

PR #49, opened 2026-01-14
chore: configure devcontainers for improved DX

radiantBear:
This enables dev containers to be used with Tasktix, improving DX by allowing developers to make changes directly from the container running the dev server. That is particularly helpful when running commands like some Prisma commands that will make changes to the container that aren't copied back to the host machine. With dev containers, changes can be committed and pushed directly from the container, removing the need to copy changes back.

This also allows developers to work from the browser using GitHub Codespaces, significantly accelerating build-and-test times for developers with lightweight host machines. Also, reviewers should be able to test changes in a new branch by spinning up a Codespace instead of checking out the branch, re-installing dependencies, running migrations, and rebuilding the server.

## Changes
- [x] Added a dev container configuration
- [ ] Did not set up Cypress tests to work within the dev container (requires many extra libraries not typically included in Alpine)
- [ ] Database is not automatically configured with `npm run prisma:apply` since this overlaps with #30 

## Testing

To try using GitHub Codespaces, pick the `devcontainers` branch on GitHub, click the "Code" button, select the Codespaces tab, and click "Create codespace on devcontainers".

image

After the Codespace finishes building, you can open the terminal, click the "Ports" tab, and click the globe icon next to the URL to view the dev version of Tasktix.
image

### DIFF
--- a/scripts/setup_devcontainer.sh
+++ b/scripts/setup_devcontainer.sh
@@ -5,7 +5,7 @@
 cp -a /app/node_modules/. /workspaces/Tasktix/node_modules
 
 # Copy the user's signing key if it exists (will only be the case locally)
-if [ -f /root/.ssh/git_signing ]; then
+if [[ -f /root/.ssh/git_signing ]]; then
   git config user.signingkey /root/.ssh/git_signing.pub
 fi
 


sabri414 (approved):
This is a game changer! thanks for putting in the effort to make this happen

PR #50, opened 2026-01-16
fix: newly-created tags not linking to the item they were added to

### LINKED ISSUE #25: Tag assignment fails when tag is first created
When a new tag is created by clicking a list item's tags and entering a new tag name, the UI shows the tag is correctly added to the list item. However, that does not persist after the page is refreshed and the tag must be added a second time.

https://github.com/user-attachments/assets/51c9da65-c6ac-4739-a69a-993efa8d8c62

radiantBear:
Fixes a bug where, when creating a new tag and simultaneously linking it to a list item, the user has to hard-refresh the page and re-add the tag to the list item in order for the assignment to persist in the database.

This bug emerged because the React state was being updated to add the tag without making an API request to link the tag in the database.

## Related Issues

- Closes #25

## Changes
- [x] Fixed the bug by adding an API call to link the tag before updating React state

## Testing 

I added a Jest test to verify that:
- The API endpoint for creating a tag is hit
- The API endpoint for linking a tag to the item is hit <mark>(this was broken before)</mark>
- The tag shows up on the list item

You can also manually test that this change works:
1. Create a new list
2. Add a section to the list & an item to the section
3. Click the "tags" dropdown on the item
4. Type a new tag name, pick a color, and click the checkmark
5. Wait for the new tag to show up on the list item
6. Refresh the page & verify that the tag still shows up on the list item

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,66 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/docker-existing-docker-compose
+{
+  "name": "Tasktix Development",
+
+  // Update the 'dockerComposeFile' list if you have more compose files or use different names.
+  // The .devcontainer/docker-compose.yml file contains any overrides you need/want to make.
+  "dockerComposeFile": ["../compose.yaml", "../compose.override.yaml"],
+
+  // The 'service' property is the name of the service for the container that VS Code should
+  // use. Update this value and .devcontainer/docker-compose.yml to the real service name.
+  "service": "web",
+
+  // The optional 'workspaceFolder' property is the path VS Code should open by default when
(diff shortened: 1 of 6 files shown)


jest.setup.ts
```diff
@@ -0,0 +1,20 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
+globalThis.structuredClone = (value: unknown) =>
+  JSON.parse(JSON.stringify(value)) as unknown;
```
radiantBear: Unfortunately, JSDom doesn't provide a global `structuredClone` function. This polyfills that function to allow its use in tested code

PR #73, opened 2026-02-10
feat(api)!: migrate to industry standard auth provider

### LINKED ISSUE #22: Migrate auth provider to Secure Provider [betterAuth]
Use supabase for auth provider instead of in-house solution.

Blocks production deployment

Ryangman:
Migrates custom auth implementation to [Better-Auth](https://www.better-auth.com/). 

>[!CAUTION]
> Pulling up to this change causes breaking database changes and will result in user account loss. It should be possible to manually remediate this which will be investigated by @radiantBear. See Reviewer Notes for more details.

This PR is not ready to merge as I will need to pull in changes from #45 to fix tests broken by the User object, but the rest of the changes are functionally ready for review.

Closes #22, Unblocks #64 

## Changes

- [x] Replaced Session Management with BetterAuth Implementation
- [x] Replaced Custom Sign-in, Sign-out, and CreateAccount routes with BetterAuth implementation
- [x] Performed Database migrations to accommodate BetterAuth Requirements
- [x] Updated Profile Update backend to use BetterAuth wrappers, user and account management
- [x] Update User object model/constructor and all tests that involved them (Waiting on #45)
- [x] Added back-end support for password changes to unblock #64 
- [x] Remove all unused code with (e.g custom session management, hashing, manual database writes/reads for user account management)

## Testing Coverage

Modified Profile Update back-end tests for new methods of managing users and accounts

Removed tests and implementations for unused functions (e.g. hashing, session management), endpoints (e,g, sessions, log in/out)

Please let me know of any additional areas that need testing with these changes. (e.g. We have no component tests for sign in, sign out, create account behavior that would have verified our previous custom API)

## Reviewer Notes
- There are new environment variables you will need to set, see `.env.example`
- It may be simplest to look at this PR by directory structure (e.g `@lib` changes, api changes, test changes, front-end changes) rather than by commit
- Reviewing/merging this causes breaking database changes (see warning above). It should be possible to perform the migrations with `npm run prisma:deploy` *without* tearing down your docker volume, but legacy user accounts will be unable to login.

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -58,10 +58,3 @@ The required schema for BetterAuth can be found [here](https://www.better-auth.c
 This results in the `User` table having the fields `name`, `username`, and `displayUsername`. For our circumstances, whenever creating/updating a username, you should be sure to update both `name` and `username` to prevent them getting out of sync. In general, consider the `username` field the source of truth. Note that the `auth.api.isUsernameAvailable` only checks the `username` column. 
 
 `displayUsername` need never be considered/provided. Unless it is manually reconfigured, it will always be in sync with `username`.
-
-
-
-
-### Manual Fixes
-1) BEFORE PULLING UP: save a backup of your database
-<!-- TODO: NATE document database update steps if worthwhile -->


.env.example
```diff
@@ -3,5 +3,7 @@
 # ##########################################################
 DB_DATABASE=tasktix
 DB_PASSWORD=password
+BETTER_AUTH_SECRET=generate a random secret value, 32 char min
+BETTER_AUTH_URL=www.google.com # Base URL of App
```
radiantBear: Providing no default value for the secret is a good idea. For the URL, though, wouldn't it be fine to default to the dev URL?
```suggestion
BETTER_AUTH_URL=http://127.0.0.1:3000 # Base URL of app
```

prisma/schema.prisma
```diff
@@ -115,36 +61,136 @@ model Tag {
...
+  Orange
+  Amber
+  Yellow
+  Lime
+  Green
+  Emerald
+  Cyan
+  Blue
+  Violet
+}
```
radiantBear: Why is this defined separately from `enum Color`? It looks like they accept the same range of values

prisma/schema.prisma
```diff
@@ -1,109 +1,55 @@
...
 
 model ListSection {
   id     String @id @db.Char(16)
   listId String @db.Char(16)
   name   String @db.VarChar(64)
-
-  items Item[]
-  list  List   @relation(fields: [listId], references: [id], onDelete: Cascade, onUpdate: Cascade)
+  items  Item[]
+  list   List   @relation(fields: [listId], references: [id], onDelete: Cascade)
```
radiantBear: This initially looked like just a formatting change, but is there a reason the `onUpdate: Cascade` was removed?

prisma/schema.prisma
```diff
@@ -115,36 +61,136 @@ model Tag {
...
 
   @@id([userId, itemId])
   @@index([itemId])
 }
+
+model Account {
+  id                    String    @id
+  accountId             String    @db.Text
+  providerId            String    @db.Text
+  userId                String
```
radiantBear: `userId` has a specified database type with 191 characters everywhere else. Shouldn't it here, too?

lib/model/__tests__/session.test.ts
radiantBear: It looks like `lib/model/session.ts` was unaffected by this PR. Was that a mistake or can these tests be restored?

prisma/schema.prisma
```diff
@@ -115,36 +61,136 @@ model Tag {
...
+  User      User     @relation(fields: [userId], references: [id], onDelete: Cascade)
+
+  @@index([userId])
+}
+
+model User {
+  id              String         @id @db.VarChar(191)
+  name            String         @unique @db.VarChar(32)
+  email           String         @unique @db.VarChar(128)
+  legacyPassword  String?        @db.Char(205)
```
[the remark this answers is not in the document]
Ryangman: I left that as it should make it easier to copy the old accounts hash:salt to manually copy old accounts into the new structure like you were planning with the production version

lib/model/__tests__/session.test.ts
Ryangman: `lib/model/session.ts` was deleted by this PR, git decided to rename the object for `app/api/auth/[...all]/route.ts`

prisma/schema.prisma
Ryangman: The relation appears handles database typing all on it own, and makes it so we only don't have to update every type declaration if we need to change it.

Image

prisma/schema.prisma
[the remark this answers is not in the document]
radiantBear: 1. Yeah, that's what I meant, thanks. Also gotcha- annoying how that keeps happening.
2. Agreed that it's not worth the effort to change BetterAuth's naming conventions. I do think consistency would be good though, so I'd vote for adopting that convention for naming relationships (we can do that separately; I'll create a new issue for it).

prisma/schema.prisma
radiantBear: Oh, nice! Right now there's some inconsistency on whether foreign keys' types are explicitly stated. I agree that it's good to not list more than needed, but we should standardize on that.

lib/database/user.ts
```diff
@@ -25,7 +25,7 @@ import { prisma } from './db_connect';
 /**
  * Updates a user Color in the database
  * Custom implementation necessary because BetterAuth doesn't provide API for modifying custom Fields
- * @param user 
+ * @param user
  * @returns Boolean
```
radiantBear: These `@returns` don't provide any info that the type signature doesn't, so they can be removed

lib/model/user.ts
```diff
@@ -19,49 +19,64 @@
...
-    .string()
-    .min(3)
-    .max(32)
-    .regex(/^[a-zA-Z0-9_]*$/),
+  id: z.string().length(191),
+  name: userFormat,
+  username: userFormat,
   email: z.email().max(128),
-  password: z.string().min(10).max(128),
+  legacyPassword: z.string().min(10).max(128),
```
radiantBear: This won't be around much longer in the database and it looks like this is omitted every time the ZodUser is used anyway.
```suggestion
```

lib/session.ts
```diff
@@ -18,49 +18,21 @@
...
-  return user;
-}
-
-export async function clearUser(): Promise<boolean> {
-  const sessionId = (await cookies()).get('user')?.value;
-
-  if (!sessionId) return false;
-
-  const result = await deleteSession(sessionId);
+  if (!user) return false;
```
radiantBear: The type of `getUserById` is already `false | User`
```suggestion
```

package.json
```diff
@@ -40,6 +40,7 @@
     "react-bootstrap-icons": "^1.11.6",
     "react-dom": "19.2.3",
     "sharp": "^0.34.3",
+    "tsx": "^4.21.0",
```
radiantBear: Please move this to `devDependencies`

lib/database/user.ts
```diff
@@ -57,21 +50,14 @@ export async function getUserByUsername(
   return result ?? false;
 }
 
+/**
+ * Returns a User object associated with email
+ * Custom Implementation as querying an email is not supported by betterauth
+ * @param email
+ * @returns
```
radiantBear:
```suggestion
```

lib/database/user.ts
```diff
@@ -22,22 +22,15 @@ import User from '@/lib/model/user';
...
-  }
-
-  return true;
-}
-
-export async function updateUser(user: User): Promise<boolean> {
+/**
+ * Updates a user Color in the database
+ * Custom implementation necessary because BetterAuth doesn't provide API for modifying custom Fields
+ * @param user
```
radiantBear: Please provide a description next to each `@param` (line +56 also)

docs/auth.md
```diff
@@ -0,0 +1,66 @@
+# Authentication and Authorization
+
+The Tasktix project uses [Better-Auth](https://www.better-auth.com/) to provide authentication services for the application. BetterAuth was chosen over supabase or other alternatives because our database was setup as mysql, and additionaly migrating to postgres was deemed unnecessary. Better-Auth natively supports oauth and two factor which are left as future updates.
```
radiantBear: Could you please correct the capitalization of "Supabase", "MySQL", "OAuth", etc?

app/api/user/[id]/route.test.ts
```diff
@@ -145,55 +159,130 @@ describe('PATCH', () => {
...
+        body: JSON.stringify({
+          newPassword: 'new_password',
+          oldPassword: 'old_password'
+        })
+      }),
+      { params: Promise.resolve({ id: MOCK_USER.id }) }
+    );
+
+    expect(response.status).toBe(200);
+    expect(auth.api.changePassword).toHaveBeenCalled();
```
Comment (radiantBear) [inline app/api/user/[id]/route.test.ts]: Other tests assert what the mocked function was called with. Shouldn't this one too?

app/api/user/[id]/route.test.ts
```diff
@@ -208,11 +297,12 @@ describe('PATCH', () => {
     );
 
     expect(response.status).toBe(400);
-    expect(updateUser).not.toHaveBeenCalled();
+    expect(auth.api.updateUser).not.toHaveBeenCalled();
   });
 
   test('Rejects invalid email updates', async () => {
     vi.mocked(getUser).mockResolvedValue(MOCK_USER);
+    expect(auth.api.changeEmail).not.toHaveBeenCalled();
```
Comment (radiantBear) [inline app/api/user/[id]/route.test.ts]: This assertion should be unnecessary since `vi.resetAllMocks()` is called in the `beforeEach` hook

prisma/migrations/20260219181711_auth_refactor/migration.sql
```diff
@@ -0,0 +1,173 @@
...
+
+-- AlterTable
+ALTER TABLE `User` DROP PRIMARY KEY,
+    DROP COLUMN `dateCreated`,
+    DROP COLUMN `dateSignedIn`,
+    ADD COLUMN `createdAt` DATETIME(0) NOT NULL DEFAULT CURRENT_TIMESTAMP(0),
+    ADD COLUMN `displayUsername` VARCHAR(32) NULL,
+    ADD COLUMN `emailVerified` BOOLEAN NOT NULL DEFAULT false,
+    ADD COLUMN `image` TEXT NULL,
+    ADD COLUMN `name` VARCHAR(32) NOT NULL,
```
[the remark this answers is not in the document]
Ryangman: instead renamed the original username column (unique and nonnull) to name, then added nullable username and displayUsername and copied name later

prisma/schema.prisma
```diff
@@ -82,8 +82,8 @@ model ListMember {
 }
 
 model ItemAssignee {
-  userId String
-  itemId String
+  userId String @db.VarChar(191)
```
radiantBear: It looks like `userId` foreign keys now sometimes have a database type specified (e.g. here) and sometimes don't (e.g. line 99). If it doesn't require any migrations, could you please standardize those?

radiantBear (dismissed):
Looks good! Couple minor critiques left, but I think this is ready to go! I'll work on testing actual migration to this now.

radiantBear (approved):

docs/auth.md
```diff
@@ -0,0 +1,66 @@
+# Authentication and Authorization
+
+The Tasktix project uses [Better-Auth](https://www.better-auth.com/) to provide authentication services for the application. Better-Auth was chosen over Supabase or other alternatives because our database was setup as MySQL, and additionaly migrating to Postgres was deemed unnecessary. Better-Auth natively supports OAuth and two factor which are left as future updates.   
+
+>[!CAUTION]
+> Pulling up a previous Tasktix version to adopt Better-Auth includes breaking database changes, and will result in loss of User Accounts. This can be manually remediated by peforming the steps in the Database Updates section.
```
radiantBear:
```suggestion
> [!CAUTION]
>
> Pulling up a previous Tasktix version to adopt Better-Auth includes breaking database changes that require manual resolution. It also contains unavoidable password hashing changes and will result users losing the ability to sign in with their current account.
```

radiantBear (approved):

lilyvesstible (approved):
Looks good to merge!

Issue #56, opened 2026-01-20
Notify user when offline that changes will be synced


radiantBear:
Needs to include caching behavior (LocalStorage?) to save requests & retry if failed/offline

PR #78, opened 2026-02-17
refactor(ui): move state to the top level for SSE updates

radiantBear:
Refactoring the list item state to the `<List />` level will enable updates via Server-Sent Events. If the state is left at the `<ListItem />` level, every list item would need a separate connection to the server to listen for state updates for that specific item. This simply does not scale, so instead, we need to make a single connection at the `<List />` level and bubble state changes down to the modified items. This refactor sets up that type of state propagation, setting the stage for live updates via SSE.

## Related Issues
- Related to #56 

## Changes

- [x] Relocated state to the `<List />` level
- [x] Added tests for a large chunk of affected state changes

### Tests added

- Most affected state changes are tested by integration tests in `List.test.tsx`

### Tests needed

- Tests skipped for due date changes because #80 will likely change the interface for interacting with the component in tests
- Tests skipped for reordering list items via dragging because #47 will likely change this interface
- Tests skipped for API calls' error path since #79 will remove these
- Tests skipped for filtering logic since this will be replaced with #53 

## Reviewer Notes

Are comments & naming sufficient to easily understand how state propagates?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -79,3 +79,4 @@ jobs:
           platforms: linux/amd64,linux/arm64
           push: true
           tags: ${{ steps.meta-deploy.outputs.tags }}
+          file: Dockerfile.deploy


lilyvesstible (approved):
Everything looks in place! I'm glad you did this, I did notice that the list item params were a bit cluttered.

radiantBear:
> 3. Change setting to not track time
> 4. Try to add new item (without time) -> Request Body Failed
> 5. Switch setting back to tracking time
> 6. Try to add new item (with time) -> Request Succeeds
> 
> Please investigate this on your machine and see if this is just something on my side

Thanks for the robust validation of this PR! This bug is present in production as well; the Zod rule for the `POST /item` endpoint is too strict. 

I just created #87 to address this and will make that my next unit of work.

PR #85, opened 2026-02-20
feat(api): integrate auth with leaked password repository

Ryangman:
Integrates Authentication Provider with [HaveIBeenPwned ](https://haveibeenpwned.com/) leaked password repository to prevent users using otherwise *secure* passwords that have been compromised in data leaks

## Changes

- [x] Installed HaveIBeenPwned plugin
- [x] Added environment variable to configure enablement of HaveIBeenPwned 
- [x] Fixed broken test / Added new test

### Tests added

- Added e2e test for toast notification content

### Tests Broken

- Existing unit tests are not broken because we mock the auth provider. In interaction with live database/auth server many unit tests would fail because the password they choose is not secure (e,g, `new_password`)

### DIFF
--- a/docs/self_hosting.md
+++ b/docs/self_hosting.md
@@ -37,7 +37,10 @@ these:
 Pod=tasktix.pod
 Image=ghcr.io/tasktix/tasktix-web
 
+Environment=BETTER_AUTH_URL=<PATH_TO_TASKTIX_ROOT>
+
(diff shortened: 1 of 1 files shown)


radiantBear (dismissed):
Looks pretty good! I think it'd be good to make it more clear to users why the password was rejected when the frontend shows it as valid, though.

sabri414 (approved):
Great security enhancement and clean implementation. The feature flag + tests look solid. Looks good to me

PR #88, opened 2026-02-20
docs: update self-hosting docs for BetterAuth environment variables

radiantBear:
BetterAuth requires an additional environment variable and secret to be configured when deploying Tasktix. This updates the documentation to explain how these should be set to securely deploy Tasktix.

## Related Issues
- Related to #22


sabri414 (approved):
the changes are correct, focused, and improve the docs without affecting code behavior.

PR #89, opened 2026-02-20
fix(api): accept new items w/o expected time when disabled on list

### LINKED ISSUE #87: Items can't be added to lists without time tracking
_See https://github.com/Tasktix/Tasktix/pull/78#pullrequestreview-3830365363 for details_

radiantBear:
Fixes a bug where users can't create new list items without specifying the expected time to complete the item. When time tracking is disabled, `undefined` and `null` times should both be valid.

## Related Issues

- Closes #87 

## Changes

- [x] Fixed Zod rule causing valid updates to be rejected
- [x] Added regression test for bug
- [x] Added docblock for API route
- [x] Added other tests for API route

## Testing Coverage

All functionality in the affected API route is now unit tested.


sabri414 (approved):
looks good, Given that 1970 wouldn’t realistically occur through the UI, I’m good with this as-is

PR #81, opened 2026-02-19
feat: push events for list metadata updates

radiantBear:
This PR starts a minimal implementation of Server-Sent Events for updating actual list state based on other users' changes.

## Related Issues

- Closes #55 
  - More work to finalize SSE, but this finalizes the interface
- Related to #13 

## Changes

- [x] Updated SSE lib functions to expect `ListAction`s
  - [x] Tested SSE lib functions
- [x] Added client-side SSE lib function to consolidate logic
- [x] Added SSE broadcasts for list metadata changes
- [x] Tested SSE-enabled API endpoint

PR #90, opened 2026-02-21
refactor(api): use better function names for getting members

radiantBear:
This PR renames 2 functions that fetch list members that currently claim to have something to do with assignees. Members are people with access to a list; assignees are people who are supposed to complete a specific item. The terms should not be used interchangeably.

## Reviewer Notes

DeepSource documentation problems should be resolved when #81 is merged; the comments shouldn't be duplicated here.


A-man37 (approved):
Looks good!

Issue #94, opened 2026-02-25
Integrate With Github Projects


Ryangman:
*Further sub-issues will be created as integration progresses*

PR #97, opened 2026-02-26
feat: add OAuth login support for GitHub

### LINKED ISSUE #95: Add Github OAuth

Ryangman:
Adds OAuth Login Support for Github

## Related Issues

Closes #95 
Unblocks #94 

## Changes
- [x] Added Oauth Login Handler, Buttons on Sign in/Sign up Page
- [x] Changed Profile Icon displays to use Image if it exists 
- [x] Fixed use of `username` `name` and `displayUsername` fields across app

## Testing Coverage

Requires setting up a Tasktix Github App via Github GUI and set environment variables provided by that Github App. I can likely set this up within Tasktix Org but will need @radiantBear for perms.
### Tests added

- Added Tests for Oauth Handler logic
- Added Component Test for OAuth component
- 100% coverage on new lines/files

- Did not add Component tests for Sign in/Sign up components that were untested before this PR

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -10,26 +10,21 @@ title: Tasktix System Architecture
 flowchart LR
   c[Client]
   subgraph AWS EC2
-
+    direction BT
     s[Web server]
-
     d@{shape: cyl, label: Database}
-
-  end
-  subgraph AWS Lambda
-    w[Websocket server]
+    a[BetterAuth Authentication]
   end
 
   subgraph External services
     g@{shape: processes, label: Services (e.g. GitHub)}
-    a[Auth provider]
   end
 
-  c <-->|HTTP| s <-->|MySQL| d
-  s <-->|WS| w <-->|WS| c
+  c <--->|HTTP| s ---->|SSE| c
+  s <-->|MySQL| d
   g -->|HTTP| s
-  c -->|HTTP| a
   a <-->|HTTP| s
+  a <-->|HTTP| g
 ```
 
 Some key notes about the architecture design:
@@ -37,11 +32,12 @@ Some key notes about the architecture design:
 - The web server is responsible for performing all persistent state updates after
   authentication & authorization checks. These state updates include starting/ending user
   sessions and persisting user-entered data.
-- The websocket server is responsible for letting all clients know when a state change has
-  occurred for the list they are viewing (i.e. from another user's interaction). This is
-  separated from the main web server because Next.js doesn't natively support websockets.
-  Also, AWS Lambda functions scale to zero when not in use, helping reduce costs while our
-  user base remains small.
+- Tasktix utilizes Server-Sent Events to push state changes to connected clients who are
+  viewing a list that is shared by multiple clients (e.g. multiple User accounts or a
+  single User on a mobile device and desktop)
+- Better Auth Authentication operates similar to a microservice, but contained within the
+  Tasktix web server. It has additional communication with the external services such as
+  GitHub to provide OAuth/Social Sign On.
 - Tasktix can communicate with other services (e.g. GitHub) over HTTP via webhooks those
   services expose.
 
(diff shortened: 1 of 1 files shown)


app/signIn/components/OAuth.tsx
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * OAuth component that provides GitHub authentication options for user sign-in.
+ *
+ * Displays a divider section with an "OR" label followed by a GitHub sign-in button.
+ * Uses React's `startTransition` to handle the OAuth flow without blocking the UI.
+ *
+ * @component
+ * @param {OAuthControllers} props - The OAuth controller props
+ * @param {Function} props.setLoggedInUser - Callback function to set the logged-in user state
+ * @param {NextRouter} props.router - Next.js router instance for navigation after authentication
```
radiantBear: Please remove the types from this doc block. Types are provided by TS annotations, so duplicating them here increases the risk of documentation falling out of sync with code.

app/signIn/oauth.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+export async function handleOauth(
+  provider: string,
+  controllers: OAuthControllers
+) {
+  // Error Early if provider is not supported
+  if (!SUPPORTED_PROVIDERS.includes(provider)) {
+    return;
+  }
```
radiantBear: Instead of checking at run-time whether the provider is valid, I think it would be better to catch this using types.

To do so, you would first change:
```diff
- const SUPPORTED_PROVIDERS = ['github'];
+ type Provider = 'github';
```

Then,
```suggestion
  provider: Provider,
  controllers: OAuthControllers
) {
```

When more providers are added, they can be included in the type signature like this:
```diff
- type Provider = 'github';
+ type Provider = 'github' | 'google';
```

`SUPPORTED_PROVIDERS` and thus `Provider` are only used in 1 place, so it should also be fine to inline the type instead of declaring it above.

app/signIn/oauth.ts
```diff
@@ -0,0 +1,78 @@
...
+      provider
+    },
+    {
+      onSuccess: (ctx: SuccessContext<{ User: User }>) => {
+        controllers.setLoggedInUser(ctx.data.User);
+        controllers.router.push('/list');
+      },
+      onError: ctx => {
+        addToast({ title: ctx.error.message, color: 'danger' });
+      }
```
radiantBear: This could be a later refactoring, but instead of writing this here, in `SignIn.tsx`, and in `SignUp.tsx`, it would probably be better for the components to accept `onSuccess` and `onError` props and define these handlers in a single place.

app/signIn/oauth.ts
```diff
@@ -0,0 +1,78 @@
...
+ * @example
+ * ```typescript
+ * await handleOauth('github', {
+ *   setLoggedInUser: (user) => setUser(user),
+ *   router: useRouter()
+ * });
+ * ```
+ */
+
+export async function handleOauth(
```
radiantBear:
```suggestion
export async function handleOAuth(
```

app/signIn/components/SignIn.tsx
```diff
@@ -78,11 +79,13 @@ export default function SignIn() {
         variant='bordered'
         onValueChange={handlePasswordInput}
       />
-      <div className='flex justify-center mt-6'>
+
+      <div className='flex justify-center my-4'>
```
radiantBear: I agree, decreasing the margin between the inputs and the button is good. I think it would be better to let the "OR" divider handle the bottom margin, though. If OAuth were conditionally rendered, putting the margin here could cause unintended spacing changes to the bottom of the card.
```suggestion
      <div className='flex justify-center mt-4'>
```

app/signIn/components/SignUp.tsx
radiantBear: Don't forget to adjust this margin to match `SignIn`!
```suggestion
      <div className='flex justify-center mt-4'>
```

app/signIn/components/OAuth.tsx
```diff
@@ -33,12 +32,20 @@ import { handleOAuth, OAuthControllers } from '../oauth';
...
  * <OAuth setLoggedInUser={setUser} router={router} />
  */
-export default function OAuth({ setLoggedInUser, router }: OAuthControllers) {
+export default function OAuth({
+  setLoggedInUser,
+  oauthConfig
+}: Readonly<OAuthControllers>) {
+  if (!oauthConfig?.githubEnabled) {
+    /* Required to appease TypeScript skipcq JS-0045*/
+    return <></>;
```
radiantBear: This can be replaced with `return null` for future reference

PR #99, opened 2026-02-28
feat(ui): mobile sidebar drawer and responsive safeguards

### LINKED ISSUE #54: Make Tasktix WCAG 2.1 compliant

### LINKED ISSUE #103: Make Tasktix mobile friendly
Things to do for mobile friendliness:

- [x] Fix the footer, make it at the bottom of the page and not follow the user as they scroll.
- [x] Make the gear button on the right of the filter bar 
(seperate): fix the toggle screen from light and dark mode to be able to change the theme on click instead of on hover 
- [x] Shrink the text box for the name of the task a little bit and only allow changes to be made inside the details menu for mobile screens only
- [ ] `getbyrole` button then have an event that clicks the button, then get the sidebar drawer. we don't need to use dialog here 
- [x] Make List.complete-flow.test.tsx an end 2 end test using cypress. make some of the test components inside complete-flow inside list.cy.ts
- [x] When a user tries to to edit the name of a task there is a green check box that shows up inside the rest of the UI that make it hard for a user to make those changes. we want the green box to appear in a place where the user can click on it

sabri414:
Responsive list layout: mobile sidebar drawer

- Hides sidebar on small screens; keeps desktop unchanged (md+)
- Adds slide-over drawer for sidebar nav on mobile
- Adds accessibility: focus management, Esc to close, overlay click to close
- Adds overflow safeguards to prevent horizontal scrolling
- Notes: local lint/typecheck wasn’t run due to environment issues; please verify in CI/Codespaces


app/providers.tsx
radiantBear, line 1: Please revert this change to avoid cluttering `git blame`

components/List/__tests__/List.test.tsx
```diff
@@ -386,6 +386,8 @@ describe('Tag changes', () => {
   test('Allows tags to be deleted', async () => {
     const user = userEvent.setup();
 
+    vi.spyOn(globalThis, 'confirm').mockImplementation(() => true);
+
```
radiantBear: `confirm()` should not be in use anymore; #127 provides a replacement that fits the style of our app

components/List/__tests__/List.test.tsx
```diff
@@ -1250,8 +1252,8 @@ describe('ListItem state propagation', () => {
       expect(popoverElement).toHaveAccessibleName('Elapsed 00:02');
     });
 
-    // TODO: Need tests for timers actually incrementing after @testing-library/user-event
-    // supports Vitest fake timers:
+    // Timer increment behavior still needs dedicated coverage once @testing-library/user-event
+    // fully supports Vitest fake timers:
```
radiantBear: Please revert this change to avoid cluttering `git blame`; it essentially says the same thing as before

package.json
```diff
@@ -20,12 +20,12 @@
...
-    "test:cy": "CYPRESS_EXIT=0; npm run start:test -- --detach && npm run testdb:reset; cypress run || CYPRESS_EXIT=$?; npm run stop:test && exit $CYPRESS_EXIT",
+    "test:cy": "CYPRESS_EXIT=0; npm run start:test -- --detach --wait && npm run testdb:reset; cypress run || CYPRESS_EXIT=$?; npm run stop:test && exit $CYPRESS_EXIT",
     "start:test": "cross-env DB_PASSWORD=password DB_DATABASE=tasktix_test BETTER_AUTH_URL=\"http://localhost:3000\" BETTER_AUTH_SECRET=\"CGWdsHolhwwL4Le08c7yvJKU01x3K0kRkoiYoWStseM=\" docker compose -f compose.yaml -f compose.test.yaml up --build",
     "stop:test": "docker compose -f compose.yaml -f compose.test.yaml down --volumes",
-    "testdb:reset": "docker compose exec deploy npm run prisma:reset -- --force",
-    "testdb:seed": "docker compose exec deploy npm run prisma:seed",
-    "cy:open": "cypress open"
+    "testdb:reset": "docker compose -f compose.yaml -f compose.test.yaml exec -T deploy npm run prisma:reset -- --force",
+    "testdb:seed": "docker compose -f compose.yaml -f compose.test.yaml exec -T deploy npm run prisma:seed",
+    "cy:open": "npm run start:test -- --detach --wait && cypress open"
```
radiantBear: As explained in `CONTRIBUTING.md`, `cy:open` should not start the test server. This is intentionally left to another command to improve visibility of server `console.error` messages while interactively running end-to-end tests. If automatic, non-interactive end-to-end test runs are needed, `test:cy` should be used instead.

```suggestion
    "cy:open": "cypress open"
```

package.json
```diff
@@ -20,12 +20,12 @@
     "prisma:reset": "prisma migrate reset --skip-seed",
     "prisma:seed": "prisma db seed",
     "test:vi": "vitest",
-    "test:cy": "CYPRESS_EXIT=0; npm run start:test -- --detach && npm run testdb:reset; cypress run || CYPRESS_EXIT=$?; npm run stop:test && exit $CYPRESS_EXIT",
+    "test:cy": "CYPRESS_EXIT=0; npm run start:test -- --detach --wait && npm run testdb:reset; cypress run || CYPRESS_EXIT=$?; npm run stop:test && exit $CYPRESS_EXIT",
```
radiantBear: `--wait` implies `--detach`
```suggestion
    "test:cy": "CYPRESS_EXIT=0; npm run start:test -- --wait && npm run testdb:reset; cypress run || CYPRESS_EXIT=$?; npm run stop:test && exit $CYPRESS_EXIT",
```

package.json
```diff
@@ -20,12 +20,12 @@
...
     "start:test": "cross-env DB_PASSWORD=password DB_DATABASE=tasktix_test BETTER_AUTH_URL=\"http://localhost:3000\" BETTER_AUTH_SECRET=\"CGWdsHolhwwL4Le08c7yvJKU01x3K0kRkoiYoWStseM=\" docker compose -f compose.yaml -f compose.test.yaml up --build",
     "stop:test": "docker compose -f compose.yaml -f compose.test.yaml down --volumes",
-    "testdb:reset": "docker compose exec deploy npm run prisma:reset -- --force",
-    "testdb:seed": "docker compose exec deploy npm run prisma:seed",
-    "cy:open": "cypress open"
+    "testdb:reset": "docker compose -f compose.yaml -f compose.test.yaml exec -T deploy npm run prisma:reset -- --force",
+    "testdb:seed": "docker compose -f compose.yaml -f compose.test.yaml exec -T deploy npm run prisma:seed",
```
radiantBear: Need to fact-check this, but I believe it's beneficial to allocate a tty for better error message visibility
```suggestion
    "testdb:reset": "docker compose -f compose.yaml -f compose.test.yaml exec deploy npm run prisma:reset -- --force",
    "testdb:seed": "docker compose -f compose.yaml -f compose.test.yaml exec deploy npm run prisma:seed",
```

components/ListItem/__tests__/ListItem.handlers.test.tsx
radiantBear, line 1: Why does this file exist? `.coverage.test.tsx` isn't part of our naming conventions

lilyvesstible (dismissed):

app/body.tsx
```diff
@@ -76,11 +76,11 @@ export default function Body({ children }: Readonly<{ children: ReactNode }>) {
         </NavbarContent>
       </Navbar>
 
-      {children}
+      <main className='flex-1 min-h-0'>{children}</main>
```
radiantBear: It's up to each `<Page>` to add the `<main>` element. Please remove it here.

components/ListItem/ListItem.tsx
```diff
@@ -192,17 +194,17 @@ export default function ListItem({
   }, []);
 
   function _stopRunning() {
-    if (timer.current) {
+    if (timer.current !== null) {
       clearTimeout(timer.current);
-      timer.current = undefined;
+      timer.current = null;
```
radiantBear: Please revert these changes; the timers are fairly brittle and are known to currently work. Unless they're fully overhauled to be more comprehensible, reviewing changes is very difficult.

components/ListItem/ListItem.tsx
```diff
@@ -225,22 +227,31 @@ export default function ListItem({
...
-                  aria-label='Item name'
-                  className='shrink'
-                  updateValue={itemHandlers.setName}
-                  value={item.name}
-                  variant='underlined'
-                />
-              </span>
+              <>
+                <span
+                  className={`min-w-0 text-sm text-foreground truncate md:hidden ${hasDueDates ? '' : 'mt-1'}`}
```
radiantBear: `min-w` here probably shouldn't be `0`

components/ListItem/ListItem.tsx
```diff
@@ -280,7 +291,7 @@ export default function ListItem({
           )}
         </span>
       </span>
-      <span className='flex gap-4 items-center justify-between w-3/5'>
+      <span className='flex gap-4 items-center justify-between w-3/5 max-w-fit lg:max-w-full'>
```
radiantBear: What's going on with `max-w` here?

radiantBear (changes requested):
Thanks Sabri, this is a great start! I still need to review a few things that were significantly overhauled, but wanted to get you some initial thoughts.

Remaining to review:
- [ ] All tests
- [ ] Sidebar component
- [ ] Theme switcher component

sabri414:
@radiantBear I addressed the requested changes. I removed the extra <main> ownership from Body, added a safe ThemeSwitcher fallback for environments without matchMedia, aligned the package scripts with the review suggestions, reverted the unnecessary timer churn, narrowed the ListItem responsive changes, renamed the extra ListItem test file to match repo conventions, and trimmed the mobile drawer/sidebar changes to the minimum needed behavior. I also ran typecheck, targeted Vitest, and ESLint on the touched files, and they all passed. Ready for another look when you are.

global.d.ts
radiantBear, line 1: This file should not be necessary; the issue with your editor reporting an error for importing globals.css is not a problem with the repo but a problem with your local environment.

app/body.test.tsx
```diff
@@ -61,6 +73,31 @@ const MOCK_OAUTH_CONFIG = {
...
+          <main className='grow'>child</main>
+        </Body>
+      </HeroUIProvider>
+    );
+
+    const main = screen.getByRole('main');
+
+    expect(screen.getAllByRole('main')).toHaveLength(1);
+    expect(main.parentElement).toHaveClass('flex', 'min-h-dvh', 'flex-col');
+    expect(main).toHaveClass('grow');
```
radiantBear: `<main>` is not part of the SUT and thus should not have assertions tied to it.

```suggestion
    expect(main.parentElement).toHaveClass('flex', 'min-h-dvh', 'flex-col');
```

radiantBear (commented):
Thanks Sabri! Here's some follow-up feedback on what I commented about before.

You made some pretty substantial further changes to `ThemeSwitcher` and `Sidebar` - anything I should know about why those were made? I'll review them when I have some more time.

components/ListItem/__tests__/ListItem.handlers.test.tsx
radiantBear: It's still not clear to me why this was separated from ListItem.test.tsx. Can you please explain?

sabri414:
> Thanks Sabri! Here's some follow-up feedback on what I commented about before.
> 
> You made some pretty substantial further changes to `ThemeSwitcher` and `Sidebar` - anything I should know about why those were made? I'll review them when I have some more time.

Yes. There were two different reasons, and I already trimmed both back to the smallest version I could justify.
For ThemeSwitcher.tsx, the bigger-looking change was mostly because the new UX split desktop and mobile behavior:
- Desktop: button click toggles light/dark directly, hover opens the full menu.
- Mobile: button opens the menu directly.
Once that behavior existed, it needed a safe device check. The important production fix was making matchMedia optional instead of assuming globalThis.matchMedia exists. Without that guard, SSR-like environments, tests, or limited browsers can throw. The current code falls back to non-mobile behavior if matchMedia is unavailable, and ThemeSwitcher.test.tsx now covers that real fallback case.

For Sidebar.tsx, the only functional reason for touching it was the mobile drawer. The drawer lives in layoutClient.tsx, and Sidebar needed just enough surface area to work in both places:
- className so the same sidebar can render differently in desktop vs drawer containers.
- onNavigate so tapping a link inside the mobile drawer can close it.
I had briefly added more scaffolding around focus/timers there, but I removed that in the cleanup pass because it wasn’t necessary. The current Sidebar diff is intentionally small: no extra drawer logic inside Sidebar, no custom focus management, just the props needed for the drawer integration plus the existing autoFocus input behavior.

global.d.ts
sabri414: I see, I brought it up in our meeting, and it seemed to persist on my local machine. I'll go ahead and remove it

tsconfig.json
```diff
@@ -25,6 +25,7 @@
   },
   "include": [
     "next-env.d.ts",
+    "**/*.d.ts",
```
radiantBear: With `global.d.ts` removed, this should be too

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
+      </Listbox>
+    </PopoverContent>
+  );
+
+  return (
+    <div ref={setContainer} className='flex items-center'>
+      <Popover
+        isOpen={isOpen}
+        placement='bottom-end'
+        portalContainer={portalEl ?? undefined}
```
radiantBear: This seems like improper use of `useState`. Iirc, this is exactly what `useRef` is intended for and you can use `containerRef` for this.

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
-  return (
-    <Popover isOpen={isOpen}>
-      <PopoverTrigger
-        onMouseLeave={() => handleMouse(false)}
-        onMouseOver={() => handleMouse(true)}
+  useEffect(() => {
+    const mediaQuery =
+      typeof globalThis.matchMedia === 'function'
+        ? globalThis.matchMedia('(hover: none), (pointer: coarse)')
+        : null;
```
radiantBear: Making this optional just so that tests will pass is an anti pattern. According to [caniuse.com](https://caniuse.com/?search=matchMedia), `window.matchMedia` is available for essentially all users. Instead of making it optional in the code, tests need to mock it if `jsdom` doesn't already provide a mock.

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
+
+    // Keep hover-only behavior disabled on touch-oriented devices.
+    function updateIsMobile(event?: MediaQueryListEvent) {
+      setIsMobile(event?.matches ?? mediaQuery?.matches ?? false);
+      setIsOpen(false);
+    }
+
+    updateIsMobile();
+    mediaQuery?.addEventListener('change', updateIsMobile);
+
```
radiantBear: This function violates SRP.
```suggestion
    setIsMobile(mediaQuery?.matches ?? false);
    
    mediaQuery?.addEventListener('change', event => {
      setIsMobile(event?.matches ?? false);
      setIsOpen(false);
    });

```

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
+    return () => {
+      mediaQuery?.removeEventListener('change', updateIsMobile);
+
+      if (closeTimeoutRef.current) {
+        clearTimeout(closeTimeoutRef.current);
+      }
+    };
+  }, []);
```
radiantBear, line 74: This effect is doing a lot of things and detracting from this component being declarative. I think it'd be nice to refactor it into a custom `useIsMobile` hook and separate `useEffect` that's solely responsible for cleaning up the `closeTimeoutRef`.

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
+  // Toggle the menu on mobile, but keep desktop click behavior as a quick theme swap.
+  function handlePress() {
+    if (isMobile) {
+      setIsOpen(open => !open);
+
+      return;
+    }
+
+    setTheme(nextTheme);
+  }
```
radiantBear: This isn't working for desktop clicks

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
+        </ListboxItem>
+        <ListboxItem key='dark' startContent={<MoonFill />}>
+          Dark
+        </ListboxItem>
+        <ListboxItem key='system' startContent={<Display />}>
+          System
+        </ListboxItem>
+      </Listbox>
+    </PopoverContent>
+  );
```
radiantBear: Please inline this variable; it's hard to read when the component tree is split in two

components/ThemeSwitcher.tsx
```diff
@@ -16,91 +16,198 @@
...
+      onMouseLeave={() => {
+        if (!isMobile) {
+          scheduleClose();
+        }
+      }}
+      onPress={handlePress}
+    >
+      {activeTheme === 'dark' ? <SunFill /> : <MoonFill />}
+    </Button>
+  );
```
radiantBear: Please inline this variable or move it to its own component. It's difficult to read when the returned component tree is split up like this.

app/list/layoutClient.tsx
```diff
@@ -30,17 +39,66 @@ export default function LayoutClient({
...
+          {children}
+        </div>
       </div>
+      <Drawer
+        hideCloseButton
+        classNames={{
+          backdrop: 'md:hidden',
+          base: 'md:hidden',
+          wrapper: 'md:hidden'
+        }}
```
radiantBear: Why is this custom hiding being used? It doesn't do anything, but I assume it's supposed to accomplish the same as `backdrop='transparent'` which would be more declarative.

app/list/layoutClient.tsx
```diff
@@ -30,17 +39,66 @@ export default function LayoutClient({
...
+        isOpen={isDrawerOpen}
+        placement='left'
+        scrollBehavior='inside'
```
radiantBear: This is the default behavior and doesn't need to be specified
```suggestion
```

app/list/layoutClient.tsx
```diff
@@ -30,17 +39,66 @@ export default function LayoutClient({
...
+        <DrawerContent
+          className='flex h-dvh max-h-dvh w-72 max-w-[85vw] flex-col rounded-none border-r border-content3 p-4 shadow-2xl'
+          id='mobile-sidebar-drawer'
+        >
+          <DrawerHeader className='px-0 pb-2 pt-0 text-sm font-semibold'>
+            Navigation
+          </DrawerHeader>
+          <DrawerBody className='flex-1 px-0 pb-0 pt-0'>
+            <Sidebar
+              className='w-full flex-1 p-0 pr-0 shadow-none'
```
radiantBear: This is a *lot* of custom classes. HeroUI should do a pretty good job styling stuff by default; please remove all classes that don't have a profound impact on layout.

app/list/layoutClient.tsx
```diff
@@ -30,17 +39,66 @@ export default function LayoutClient({
...
+              onPress={onOpen}
+            >
+              <ListIcon size={20} />
+            </Button>
+          </header>
+          {children}
+        </div>
       </div>
+      <Drawer
+        hideCloseButton
```
radiantBear: Removing the close button would hinder usability. It doesn't look like it's in the way of anything
```suggestion
```

app/list/layoutClient.tsx
```diff
@@ -30,17 +39,66 @@ export default function LayoutClient({
   startingLists: string;
   children: ReactNode;
 }) {
+  const {
+    isOpen: isDrawerOpen,
+    onClose: closeDrawer,
+    onOpen,
+    onOpenChange
```
radiantBear: Please rename either all or none of these variables to include the word "drawer"

components/Sidebar/Sidebar.tsx
```diff
@@ -39,7 +37,15 @@ import { ListContext } from './listContext';
  *
  * @param lists The lists the user has access to
  */
-export default function Sidebar({ lists }: { lists: List[] }) {
+export default function Sidebar({
+  lists,
+  className = '',
```
radiantBear: `<Sidebar>` is only used in one place, so classnames should just be adjusted in the component instead of a mix of self-defined and passed in.
```suggestion
```

radiantBear (changes requested):

components/Sidebar/Sidebar.tsx
```diff
@@ -80,7 +80,7 @@ export default function Sidebar({
 
   return (
     <aside
-      className={`w-48 bg-transparent shadow-l-lg shadow-content4 p-4 pr-0 flex flex-col gap-4 overflow-auto ${className}`}
+      className={`bg-transparent flex flex-col gap-4 overflow-auto ${inDrawer ? 'w-full p-0 shadow-none' : 'hidden w-48 p-4 pr-0 shadow-l-lg shadow-content4 md:flex'}`}
```
radiantBear: Note to self: need to check what the classes dependent on `inDrawer` do and whether they're necessary

components/ThemeSwitcher.tsx
```diff
@@ -123,90 +110,85 @@ export default function ThemeSwitcher() {
...
-        <ListboxItem key='system' startContent={<Display />}>
-          System
-        </ListboxItem>
-      </Listbox>
-    </PopoverContent>
-  );
-
   return (
-    <div ref={setContainer} className='flex items-center'>
+    <div className='flex items-center'>
```
radiantBear: This `<div>` doesn't do anything and makes it really hard to read the diff for this full component tree
```suggestion
```

components/ThemeSwitcher.tsx
```diff
@@ -123,90 +110,85 @@ export default function ThemeSwitcher() {
...
+        <PopoverTrigger>
+          <Button
+            isIconOnly
+            aria-expanded={isOpen}
+            aria-haspopup='menu'
+            aria-label={
+              isMobile ? 'Open theme menu' : `Switch to ${nextTheme} mode`
+            }
+            variant='ghost'
+            onClick={handlePress}
```
radiantBear: `onClick` is deprecated
```suggestion
            onPress={handlePress}
```

components/ThemeSwitcher.tsx
```diff
@@ -39,43 +57,16 @@ import { Display, MoonFill, SunFill } from 'react-bootstrap-icons';
...
   const { theme, resolvedTheme, setTheme } = useTheme();
   const [isOpen, setIsOpen] = useState(false);
-  const [isMobile, setIsMobile] = useState(false);
-  const containerRef = useRef<HTMLDivElement>(null);
-  const [portalEl, setPortalEl] = useState<HTMLDivElement | null>(null);
+  const isMobile = useIsMobile();
   const closeTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
 
   const selectedTheme = theme ?? 'system';
   const selectedThemeKeys = new Set([selectedTheme]);
```
radiantBear: `selectedThemeKeys` is just a value computed from `selectedTheme`, which in turn is just computed from `theme`. Neither are independent variables. Instead of declaring them here, `new Set([theme ?? 'system'])` should be used where this is.

components/ThemeSwitcher.tsx
```diff
@@ -123,90 +110,85 @@ export default function ThemeSwitcher() {
...
+                setIsOpen(true);
+              }
+            }}
+            onMouseLeave={() => {
+              if (!isMobile) {
+                scheduleClose();
+              }
+            }}
+          >
+            {activeTheme === 'dark' ? <SunFill /> : <MoonFill />}
```
radiantBear: This is more clear and declarative. Plus, this allows `activeTheme` to be inlined above (i.e. `const nextTheme = resolvedTheme === 'dark' ? 'light' : 'dark'`
```suggestion
            {nextTheme === 'light' ? <SunFill /> : <MoonFill />}
```

components/Sidebar/Sidebar.tsx
radiantBear: These classes really shouldn't be the responsibility of the `<Sidebar>`. Adding an `inDrawer` prop breaks abstraction by tightly coupling this component's interface to the current layout implementation.
```suggestion
      className='bg-transparent flex flex-col gap-4 overflow-auto w-full'
```
Instead, where the `<Sidebar>` is used not inside the drawer, please wrap it like
```tsx
<div className='hidden w-48 p-4 pr-0 shadow-l-lg shadow-content4 md:flex'>
  <Sidebar ...
```
You can also delete the `inDrawer` prop with this change.

lilyvesstible (dismissed):

components/ThemeSwitcher.tsx
```diff
@@ -16,87 +16,171 @@
...
+      if (closeTimeoutRef.current) {
+        clearTimeout(closeTimeoutRef.current);
+      }
+    };
+  }, []);
+
+  // Apply the selected theme and close the menu after a choice is made.
+  function chooseTheme(themeKey: string) {
+    setTheme(themeKey);
+    setIsOpen(false);
```
radiantBear, line 79: I think it would be better to leave the listbox open when a theme is selected. When it's immediately closed and the `resolvedTheme` doesn't change (e.g. switching from Dark mode back to System), there is no visual indication to the user that their change was processed.

Since removing this leaves only one statement in the function, please go ahead and inline it, too.

radiantBear (commented):
Thanks Sabri! I'm almost done with all the review; just need to look through the tests now.

components/ThemeSwitcher.tsx
radiantBear: Update: let's go ahead and leave it for mobile since it would require a tap away to close the listbox otherwise. 

However, please don't close automatically the listbox on desktop - moving the mouse is sufficient for this, and leaving the listbox open provides visual confirmation that the user's input was accepted.

components/__tests__/ThemeSwitcher.test.tsx
```diff
@@ -0,0 +1,482 @@
...
+    render(
+      <HeroUIProvider disableRipple>
+        <ThemeSwitcher />
+      </HeroUIProvider>
+    );
+
+    fireEvent.click(screen.getByLabelText('Switch to dark mode'));
+
+    expect(setTheme).toHaveBeenCalledWith('dark');
+  });
```
radiantBear: This test should be failing; clicking the button on desktop does not change the theme on the website right now. Please fix the test and the behavior here.

components/ThemeSwitcher.tsx
radiantBear: Overruled by team

lilyvesstible (dismissed):

.env.example
```diff
@@ -10,6 +10,7 @@ BETTER_AUTH_URL=http://localhost:3000 # Base URL of APP
 
 DISABLE_LOCAL_AUTH=false
 DISABLE_PASSWORD_STRENGTH_CHECK=false
+DISABLE_AUTH_RATE_LIMIT=false
```
radiantBear, line 14: Please document why this was added (i.e. just for testing; shouldn't be done in production apps)

components/Sidebar/Sidebar.tsx
```diff
@@ -18,73 +18,40 @@
...
-import { ListContext } from './listContext';
+import CreateListModal from './CreateListModal';
 
 /**
  * Displays a sidebar with an entry for the "Today" view and an entry for each list the
  * user has access to
  *
  * @param lists The lists the user has access to
  */
+
```
radiantBear: There shouldn't be a newline between the docblock and function
```suggestion
```

components/Sidebar/Sidebar.tsx
```diff
@@ -18,73 +18,40 @@
...
-import { addToastForError } from '@/lib/error';
 
-import { ListContext } from './listContext';
+import CreateListModal from './CreateListModal';
 
 /**
  * Displays a sidebar with an entry for the "Today" view and an entry for each list the
  * user has access to
  *
  * @param lists The lists the user has access to
```
radiantBear, line 34: Please add `@param onNavigate`

cypress/e2e/list/responsive.cy.ts
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  it('opens the mobile navigation drawer and closes it after navigating to a list', () => {
+    cy.viewport('iphone-6');
+    cy.login('newUser', 'password123');
+
+    cy.request('POST', '/api/list', {
+      name: 'Mobile Cypress List',
+      color: 'Blue'
+    }).then(({ body }: { body: ApiResponseBody }) => {
```
radiantBear: Instead of making requests directly, you need to either (a) use Cypress to interact with the website to create a list, or (b) update `prisma/seed.ts` to include the new list you need. Given that this test suite isn't testing the list creation flow, (b) would be the right choice here. Please create a new user for the new list to avoid potential interference with other test suites.

cypress/e2e/list/responsive.cy.ts
radiantBear, line 1: In addition to tests for a mobile screen, there should be tests for a desktop screen that validates that the sidebar is always visible without requiring the hamburger menu to be expanded.

cypress/e2e/list/list.cy.ts
```diff
@@ -29,4 +42,46 @@ describe('Home List', () => {
...
+          cy.findByRole('checkbox').click();
+          cy.wait('@completeItem').its('response.statusCode').should('eq', 200);
+
+          cy.findByRole('checkbox').should('be.checked');
+          cy.findByText('Cypress task').should('have.class', 'line-through');
+          cy.contains(/^Completed /).should('exist');
+        });
+      });
+    });
+  });
```
radiantBear: This test has nothing to do with this PR. Please remove it.

components/SearchBar/__tests__/SearchBar.test.tsx
radiantBear, line 1: These tests have nothing to do with this PR and will be obsoleted by Andrew's PR. Please remove them.

app/list/layoutClient.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <LayoutClient startingLists='[]'>
+        <main>Content</main>
+      </LayoutClient>
+    );
+
+    const hamburger = screen.getByRole('button', {
+      name: /open navigation menu/i
+    });
+
+    expect(document.getElementById('mobile-sidebar-drawer')).toBeNull();
```
radiantBear: This assertion doesn't do anything; React doesn't expose element IDs like this.

components/__tests__/Sidebar.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+}));
+
+vi.mock('../Sidebar/CreateListModal', () => ({
+  default: ({ isOpen }: { isOpen: boolean }) =>
+    isOpen ? <div role='dialog'>Create New List</div> : null
+}));
+
+function makeList(name: string, id: string) {
+  return new List(name, 'Blue', [], [], [], false, false, false, { id });
+}
```
radiantBear: Please just inline this

app/list/layoutClient.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * @vitest-environment jsdom
+ */
+
+import { createContext } from 'react';
+import {
+  fireEvent,
```
radiantBear: Please use the `user-event` library instead of `fireEvent`

radiantBear (commented):
This completes my review for everything but `ThemeSwitcher.test.tsx`

components/SearchBar/__tests__/SearchBar.test.tsx
sabri414: That makes sense. I removed this originally to match your feedback, but re-added it after DeepSource coverage started failing because of this PR still touches `SearchBar.tsx`. I agree that the test does not belong here long-term if Andrew’s PR is replacing this path. I’m going to try removing both this test file and the `SearchBar.tsx` diff from this PR. If the responsive behavior still passes, I’ll keep it out. If the `SearchBar.tsx` change turns out to be required for this PR’s mobile layout, I’ll keep the smallest possible component change plus temporary coverage and call that out clearly.

radiantBear: We can waive the test coverage requirement for this file. Please leave the tests out.

cypress/e2e/list/responsive.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+    );
+    cy.findByRole('dialog').should('not.exist');
+  });
+
+  it('shows the desktop sidebar without opening the mobile drawer', () => {
+    cy.viewport(1024, 768);
+    cy.login('responsiveUser', 'password123');
+
+    cy.visit('/list');
+    cy.get('[aria-label="Open navigation menu"]').should('not.be.visible');
```
radiantBear: Please use the same matcher as the previous test for consistency

components/SearchBar/__tests__/SearchBar.test.tsx
sabri414: done

cypress/e2e/list/responsive.cy.ts
```diff
@@ -53,7 +53,10 @@ describe('Responsive shell', () => {
     cy.login('responsiveUser', 'password123');
 
     cy.visit('/list');
-    cy.get('[aria-label="Open navigation menu"]').should('not.be.visible');
+    cy.findByRole('button', {
+      hidden: true,
+      name: 'Open navigation menu'
+    }).should('not.be.visible');
```
radiantBear: This still isn't the same; you're checking `aria-expanded` vs. `visible`, which does not ensure consistency

components/Sidebar/Sidebar.tsx
```diff
@@ -76,28 +88,32 @@ function NavSection({
 export function NavItem({
```
lilyvesstible: Documentation might be good for this function if it's an exported function.

lilyvesstible (changes requested):

components/Sidebar/Sidebar.tsx
sabri414: I had a lapse of judgment. 'NavItem' is not used anywhere else outside of Sidebar.tsx, so I made it into a normal function

lilyvesstible (approved):

radiantBear (dismissed):

radiantBear (approved):

lilyvesstible (approved):

PR #100, opened 2026-03-01
feat: assign list member permissions via roles

### LINKED ISSUE #91: Use RBAC instead of spotty DAC

### LINKED ISSUE #112: Fix permissions gap allowing any user to update tags
_Originally posted by @Ryangman in https://github.com/Tasktix/Tasktix/pull/100#discussion_r2915196780_

radiantBear:
## Description
This PR removes the current member permissions system where users are assigned individual permissions for access to specific parts of the list. In its place, this PR adds a role-based system where each member can be assigned one of the roles stored in the database.

> [!NOTE]
>
> When updating from an older version of Tasktix, the database migrations involved in this version will make every member of every list an admin for that list.
>
> This has no practical consequence, however, because all users were able to update their own permissions in older versions of Tasktix.

This PR also addresses gaps in the old permissions system so that every possible action is tied to a role permission. The permissions a role can grant are:
- `canAddItems`
- `canUpdateItems`
- `canDeleteItems`
- `canManageTags`
- `canManageAssignees`
- `canManageMembers`
- `canUpdateList`
- `canDeleteList`

These permissions fall into 3 broad categories:
- "Add" permissions allow creating new records of the given type.
- "Update" permissions allow adding new records or changing (but not removing) existing records of the given type.
- "Delete" permissions allow deleting existing records of the given type.
- "Manage" permissions allow adding new records, changing existing records, or deleting existing records of the given type.

Sections are considered part of the list record for this permissions system (i.e. adding, updating, or removing sections requires `canUpdateList`).

### Current permissions GUI
image

### New permissions GUI
image

*Note: only the admin role is available by default. Others must be manually added to the database.*

## Related Issues

- Closes #91 
- Closes #112 
- Related to #57 

## Changes

- [x] Replaced piecemeal member permissions with comprehensive roles
  - [x] Rejects member update requests that would leave a list with no admins
- [x] Updated API routes to authorize the requestor's role before processing their request
- [x] Wrote tests for changed code

## Testing Coverage

Every API route that requires list membership includes at least 4 tests:
1. `200` success on happy path
2. `401` error if not logged in
3. `404` error if not a list member (shouldn't be able to tell that other list even exists)
4. `403` error if insufficient permission to perform the API route's action

Other tests that relied on the old `ListMember` contract have been updated to use the new role system instead.

## Reviewer Notes

Does this strategy (storing roles in the database & linking users to them) seem like the right approach for managing user permissions for lists?

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -19,12 +19,12 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules
-          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}-${{ hashFiles('prisma/schema.prisma') }}
           restore-keys: |
             ${{ runner.os }}-node-
       - name: Install dependencies if needed
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
-        run: npm ci
+        run: npm ci && npm run prisma:generate
       - name: Cache Cypress binary
         id: cache-cypress
         uses: actions/cache@v3
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: node_modules
-          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+          key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}-${{ hashFiles('prisma/schema.prisma') }}
           restore-keys: |
             ${{ runner.os }}-node-
       - name: Install dependencies if needed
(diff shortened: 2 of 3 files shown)


lib/__tests__/validate.test.ts
radiantBear: Removed unused functions & their tests

app/api/list/[id]/tag/[tagId]/route.ts
```diff
@@ -37,14 +33,16 @@ export async function PATCH(
...
+  const role = await getRoleByList(user.id, id);
 
-  if (!isMember) return ClientError.BadRequest('List not found');
+  if (!role) return ClientError.NotFound('Tag not found');
+  if (!role.canManageTags)
+    return ClientError.Forbidden('Insufficient permissions to update tag');
 
   const tag = await getTagById(tagId);
 
   // TODO: need to handle security gap: editing someone else's tag
```
Comment (Ryangman) [inline app/api/list/[id]/tag/[tagId]/route.ts]: might be worth fixing this now, seeing as we're probably not gonna find a more relevant PR to add it to.

app/api/list/route.ts
Ryangman: rename `session` to `user` to better reflect its meaning

app/api/list/[id]/tag/[tagId]/route.ts
Comment (radiantBear) [reply app/api/list/[id]/tag/[tagId]/route.ts]: This will be a reasonably-large change in its own right, especially including regression tests. I created #112 to address this.

PR #108, opened 2026-03-10
feat: add generic OIDC client for SSO

### LINKED ISSUE #107: Add generic OIDC support for custom SSO providers

radiantBear:
Adds support for Tasktix authentication via any OIDC provider, such as Authentik or Okta.

image

## Related Issues
- Closes #107 

## Changes

- [x] Added generic OIDC provider option below GitHub auth option
- [x] Allows configuring OIDC provider via environment variables

### Tests added
- Generic OIDC provider handles responses as expected
- SSO login buttons only rendered if needed environment variables are provided

### DIFF
--- a/app/signIn/oauth.ts
+++ b/app/signIn/oauth.ts
@@ -51,9 +51,20 @@ export async function handleOAuth(
   setLoggedInUser: (user: User) => void,
   oauthConfig: OAuthConfig
 ) {
+  /**
+   * Handles successful login via OAuth
+   *
+   * @param ctx The BetterAuth context for the signin
+   */
   const handleSuccess = (ctx: SuccessContext<{ User: User }>) => {
     setLoggedInUser(ctx.data.User);
   };
+
+  /**
+   * Handles errors from BetterAuth for failed login via OAuth
+   *
+   * @param ctx The BetterAuth context for the error
+   */
   const handleError = (ctx: ErrorContext) => {
     addToast({ title: ctx.error.message, color: 'danger' });
   };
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -1,12 +1,23 @@
...
 DB_PASSWORD=password
+
 BETTER_AUTH_SECRET=generate a random secret value, 32 char min
 BETTER_AUTH_URL=http://localhost:3000 # Base URL of APP
 ENABLE_PASSWORD_STRENGTH_CHECK=true
+
 GITHUB_CLIENT_ID=
 GITHUB_CLIENT_SECRET=
+
+NEXT_PUBLIC_OAUTH_PROVIDER_ID= # Will be displayed to users, so style appropriately (i.e. Title Case)
```
Ryangman:
```suggestion
NEXT_PUBLIC_OAUTH_PROVIDER_ID= #This field represents the name your provider is displayed as in the GUI, so style appropriately (e.g. Title Case)
```

Ryangman (dismissed):
lgtm

sabri414 (changes requested):
I found three regressions worth fixing before merge:

1. In Tags.tsx, the selected-tag chips are now rendered only from `tagsAvailable`, even though that prop is still optional. If `tagsAvailable` is omitted, or temporarily stale, the UI hides tags that are already present in `tags`. It would be safer to fall back to the `tags` prop when a lookup misses so renamed tags still update without breaking the existing component contract.

2. In auth.ts, `customEnabled` becomes true as soon as `NEXT_PUBLIC_OAUTH_PROVIDER_ID` and `OAUTH_CLIENT_ID` are set, but the actual OAuth config still needs enough endpoint information to function, specifically `OAUTH_DISCOVERY_URL` or `OAUTH_AUTHORIZATION_URL`. As written, the UI can advertise SSO in a partially configured environment and only fail after the user clicks it.

3. In oauth.ts, `NEXT_PUBLIC_OAUTH_SCOPES` is parsed with an unguarded `JSON.parse`. If the env var is set in a common non-JSON format like `openid profile email`, clicking the SSO button throws before the auth call is made, and the user never gets a redirect or toast. This should be validated or wrapped so misconfiguration fails in a controlled way.

PR #118, opened 2026-03-13
feat: support configuring account auth methods on profile page

### LINKED ISSUE #93: Support Deleting User Accounts and Data
The user profile page should allow a User to delete the User Account and all their associated data.

### LINKED ISSUE #113: Allow linking existing account to Github

Ryangman:
Adds the ability to configure existing accounts to be linked/unlinked to/from GitHub, also adds ability to delete account if it has a password configured

>[!Note]
> Accounts without a password (e.g. SSO) cannot delete account because account deletion requires either password or email confirmation (They don't have a password, and we don't have an email server). When password changes are added in #64, it should allow accounts without a password to set a password, using logic similar to `isGithubLinked` in this PR.

## Related Issues

Closes #113 
Closes #93 

## Changes

- [x] Added `AuthSettings` component to support linking/unlinking OAuth, deleting account, etc.
- [x] Added tests for new Component
- [x] Introduced "bug" where accounts without password cannot delete account (not an new issue as nobody could delete accounts previously)

## Testing Coverage

- Adds component tests for new `AuthSettings` component
- Adds e2e tests for delete account logic. 
   - Account deletion with stale session test is functional using `cy.wait`, but that involves busy waiting for over a minute. If fresh sessions prove problematic in future, this test can be re-enabled, but is skipped for now.


app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+    });
+  }, []);
+
+  useEffect(() => {
+    const error = queryParams.get('error');
+
+    if (error) {
+      // Error details are not given to minimize potential XSS risks
+      addToastForError(null);
+    }
```
radiantBear, line 98: I think it would be better to just use `addToast` here rather than `addToastForError` since that's intended for catching thrown errors & displaying their messages

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+  }, [queryParams]);
+
+  const isGithubLinked = accounts.some(acc => acc.providerId === 'github');
+
+  /**
+   * Attempts to a link a users Tasktix account to their Github account,
+   * redirects back to this page on success, Errors from this are handled in
+   * the useAffect above by adding toasts.
```
radiantBear:
```suggestion
   * Attempts to a link a user's Tasktix account to their Github account,
   * redirecting back to this page on success. Errors are displayed as toasts
   * by the useEffect above.
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+      await authClient.linkSocial({
+        provider: 'github',
+        callbackURL: '/profile',
+        errorCallbackURL: '/profile'
+      });
+    });
+  }
+
+  /**
+   * Attempts to unlink a users Tasktix account from Github
```
radiantBear:
```suggestion
   * Attempts to unlink a user's Tasktix account from Github
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+          }
+        }
+      );
+    });
+  }
+
+  const methods = [
+    ...(oauthConfig.githubEnabled
+      ? [
+          {
```
radiantBear: Since there's a `.filter(Boolean)` at the end of this, it should be possible to remove the spread operator & nested list (so the ternary returns an object or `undefined`).

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+            handler: isGithubLinked ? handleUnlinkGithub : handleLinkGithub,
+            isCriticalAction: false,
+            testId: 'link-to-github'
+          }
+        ]
+      : []),
+    {
+      title: 'Delete Account',
+      description: 'This action is irreversible',
+      icon: <Trash />,
```
radiantBear: The rest of Tasktix uses the filled trash can; I think consistency would be good here.

```suggestion
      icon: <TrashFill />,
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+          <div className='flex flex-col'>
+            <span className='text-base font-medium'>{method.title}</span>
+            <span className='text-sm text-gray-500'>{method.description}</span>
+          </div>
+        </div>
+        <Button
+          aria-label={method.actionLabel}
+          className={`font-medium border border-white/10 text-white ${
+            method.isCriticalAction ? 'bg-danger' : 'bg-[#27272a]'
+          }`}
```
radiantBear: > [!IMPORTANT]
> Is there a better way to do these class names?

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+        >
+          {method.actionLabel}
+        </Button>
+      </div>
+      {!isLast && <Divider />}
+    </div>
+  );
+}
+/**
+ * Renderes Modal prompting user to input their password to confirm account deletion
```
radiantBear:
```suggestion
 * Renders a modal prompting the user to input their password to confirm account deletion
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+ * @param isOpen is the modal open (from useDisclosure)
+ * @param onOpenChange how to handle changing the modals openness (from useDisclosure)
+ * @param handleDeleteAccount Handler function to attempt account deletion
+ * @returns
```
radiantBear:
```suggestion
```

app/profile/components/AuthSettings.test.tsx
```diff
@@ -0,0 +1,430 @@
...
+  'username',
+  'email@example.com',
+  false,
+  new Date(),
+  new Date(),
+  { color: 'Amber' }
+);
+
+describe('Linking/Unlinking Github', () => {
+  test('Github row not rendered Github OAuth not configured', () => {
```
radiantBear:
```suggestion
  test('Github row not rendered if Github OAuth not configured', () => {
```

app/profile/components/AuthSettings.test.tsx
```diff
@@ -0,0 +1,430 @@
...
+    expect(githubRowButton).toHaveTextContent('Disconnect');
+
+    await user.click(githubRowButton);
+
+    expect(authClient.unlinkAccount).toHaveBeenCalledWith(
+      { providerId: 'github' },
+      expect.anything()
+    );
+
+    expect(addToastForError).toHaveBeenCalledWith(exampleError.error);
```
radiantBear, line 249: I think it would be less brittle to assert on `addToast` rather than `addToastForError`

app/profile/components/AuthSettings.test.tsx
```diff
@@ -0,0 +1,430 @@
...
+    );
+    const { getByLabelText } = render(
+      <HeroUIProvider disableRipple>
+        <AuthSettings user={MOCK_USER} />
+      </HeroUIProvider>
+    );
+
+    const deleteAccountButton = getByLabelText('Delete Account');
+
+    expect(deleteAccountButton).toHaveClass('bg-danger');
```
radiantBear: This assertion doesn't really seem to fit the test case; is this standard practice or should it be removed?

app/profile/components/AuthSettings.test.tsx
```diff
@@ -0,0 +1,430 @@
...
+    expect(authClient.deleteUser).toHaveBeenCalledWith(
+      { password: 'asdf' },
+      expect.anything()
+    );
+    expect(setLoggedInUserMock).toHaveBeenCalledWith(false);
+    // eslint-disable-next-line
+    expect(routerMock.push).toHaveBeenCalledWith('/');
+  });
+
+  test('Redirects to Home after deleting account', async () => {
```
radiantBear: This test shouldn't have the same name as the previous

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+            <span className='text-sm text-gray-500'>{method.description}</span>
+          </div>
+        </div>
+        <Button
+          aria-label={method.actionLabel}
+          className={`font-medium border border-white/10 text-white ${
+            method.isCriticalAction ? 'bg-danger' : 'bg-[#27272a]'
+          }`}
+          size='sm'
+          variant='flat'
```
radiantBear: For consistency across the UI, I think it would be best to use HeroUI's default styling:

```suggestion
        <Button
          aria-label={method.actionLabel}
          color={method.isCriticalAction ? 'danger' : 'default'}
          size='sm'
          variant={method.isCriticalAction ? 'solid' : 'faded'}
```

`flat` might be better than `solid` for the critical action variant, too...

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+    // Already refactored to reasonable JSX depth skipcq: JS-0415
+    <Modal isOpen={isOpen} onOpenChange={onOpenChange}>
+      <ModalContent className='p-2'>
+        <ModalHeader className='justify-center'>
+          Confirm Account Deletion
+        </ModalHeader>
+        <ModalBody>
+          <Form className='w-full gap-4' onSubmit={handleDeleteAccount}>
+            <Input
+              isRequired
```
radiantBear: Modifying the DOM to remove `required` from the generated `<input>` element allows account deletion without password confirmation. This seems unintentional and potentially dangerous; is there a better way to require the password to be provided? If nothing else, `handleDeleteAccount` should probably perform a second validation that a password is given, though it would really be better to ensure BetterAuth does this server-side...

Also, I don't love the "required" asterisk styling since that's not used anywhere else on Tasktix since every input field in the app is required. Could that be removed?

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+      icon: <Trash />,
+      actionLabel: 'Delete Account',
+      handler: onOpen,
+      isCriticalAction: true,
+      testId: 'delete-account'
+    }
+  ].filter(Boolean);
+
+  return (
+    <div className='max-w-4xl p-6'>
```
radiantBear: This section should take up the full width (minus padding) of the card it's in

```suggestion
    <div className='p-6'>
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+      <h2 className='text-2xl font-semibold mb-6'>Account Settings</h2>
```
radiantBear: This styling doesn't align with the "Profile" header above it on the page
```suggestion
      <h2 className='text-xl p-4'>Account Settings</h2>
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+  method,
+  isLast
+}: {
+  method: AccountMethod;
+  isLast: boolean;
+}) {
+  return (
+    // Already refactored to reasonable JSX depth skipcq: JS-0415
+    <div data-testid={method.testId}>
+      <div className='flex items-center justify-between p-5 hover:bg-white/[0.02] transition-colors group'>
```
radiantBear: This version of the class is apparently preferred:
```suggestion
      <div className='flex items-center justify-between p-5 hover:bg-white/2 transition-colors group'>
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+            actionLabel: isGithubLinked ? 'Disconnect' : 'Connect',
+            handler: isGithubLinked ? handleUnlinkGithub : handleLinkGithub,
+            isCriticalAction: false,
+            testId: 'link-to-github'
+          }
+        ]
+      : []),
+    {
+      title: 'Delete Account',
+      description: 'This action is irreversible',
```
radiantBear: The current description doesn't match the grammatical style of the GitHub item
```suggestion
      description: 'Permanently delete your data,
```

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,304 @@
...
+    startTransition(async () => {
+      await authClient.deleteUser(
+        {
+          password: formData.password as string
+        },
+        {
+          onSuccess: () => {
+            setLoggedInUser(false);
+            router.push('/');
+          },
```
radiantBear, line 162: When a list is deleted and the user is redirected away from this, a "success" toast is shown to indicate that they were redirected because of successful deletion. I think it would make sense to do the same here as further visual confirmation that their account is gone.

radiantBear (changes requested):
Thank you Ryan! This is a great addition.

Can we make it so that lists are also deleted if their last member's account is removed? Also, we should explicitly handle the case of them being the last admin but not the last user of a project - maybe block account deletion until this is no longer true?

app/profile/components/AuthSettings.test.tsx
Ryangman: This assertion is included in that test because it is ensuring the delete account button is shown as critical (red button) is something worth verifying, but the amount of test setup required to make that assertion that would have to be duplicated from this test is not worth duplicating for a second test.

app/profile/components/AuthSettings.tsx
Ryangman: Because we aren't parsing the content of query params to determine what kind of error happened, this toast should show a generic "Unknown Error Occurred". I used the `addToastForError` because hitting its default unknown error case will use our generic error message and prevent duplicate copies of that string having slightly different syntax and formatting existing across the codebase.

app/profile/components/AuthSettings.tsx
Ryangman: BetterAuth is weird here, from their [docs](https://better-auth.com/docs/concepts/users-accounts#delete-user), without email verification configured, it checks for at least one of whether it was called with a correct password, or there is a fresh session token, which defaults to less than a day old. We can configure the valid session age to at a minimum 1 minute. 

Any extra verification in `handleDeleteAccount` would also be malleable on the client side. I'm not opposed to removing the asterisk, but we should discuss this as a team. It may be useful to manually handle deletion from our own backend.

app/profile/components/AuthSettings.tsx
radiantBear: That's a good point; maintaining consistency is important. I still think that it's a bit confusing to use `addToastForError` here; could you add a function named something like `addToastForDefaultError` that `addToastForError` uses if it's unable to find a message & use that directly here?

app/profile/components/AuthSettings.tsx
radiantBear: Huh, that's unfortunate. We probably should define our own backend deletion handler though - that way, we can delete lists & ensure that account deletion wouldn't leave a list with no admin.

I agree; we should discuss "required" asterisks as a team, too.

radiantBear: Instead of defining a custom backend deletion handler, perhaps [BetterAuth's hooks](https://better-auth.com/docs/concepts/hooks) are enough? `before` to validate that it's okay to delete & `after` to delete lists they're the sole user for? `ctx.context` *may* provide the password to validate that one was given, too. I haven't checked that, though.

Ryangman: They should definitely be sufficient for checking that a user isn't the only admin and deleting associated lists (though i'm surprised the sql onDelete doesn't cover that). The backend handler is more useful to prevent the "fresh session" impacting us, allowing us to always require password confirmation. I'll check to see if we get the password in that context, but i do think a single backend route might be cleaner, and easier to understand

radiantBear: Sounds good! `onDelete` doesn't cover it because the users -> lists relationship is M:N.

If the password is provided in the hook data, it might be easier to just include an `if (!password) fail;` in a beforeDelete hook, but I'm fine with the API route too if you think that's easier to understand.

app/profile/components/AuthSettings.test.tsx
```diff
@@ -306,7 +306,7 @@ describe('Account Deletion', () => {
       expect.anything()
     );
   });
-  test('Redirects to Home after deleting account', async () => {
+  test('Does not redirect to home, or delete account account deletion fails', async () => {
```
radiantBear:
```suggestion
test('Does not redirect to home, or delete account if account deletion fails', async () => {
```

radiantBear: Actually, the test implementation is the opposite of this test message. The old `'Redirects to Home after deleting account'` was accurate.

lib/database/user.ts
```diff
@@ -165,3 +165,29 @@ export async function getRoleByList(
...
+
+  for (const list of adminLists) {
+    const adminCount = await prisma.listMember.count({
+      where: { listId: list.listId, role: { name: 'Admin' } }
+    });
+
+    if (adminCount <= 1) {
+      return true;
+    }
+  }
```
radiantBear: Instead of performing `N+1` queries here, it should be possible to merge these into a single query with a subquery to select the `adminLists`.

Dockerfile.deploy
```diff
@@ -2,17 +2,20 @@ FROM node:lts-alpine
...
 RUN echo "" > "/app/node_modules/server-only/index.js"
 
 COPY scripts/test_service_entry.sh ./entry.sh
-COPY tsconfig.json prisma.config.ts ./
-COPY prisma ./prisma
 COPY lib/auth.ts ./lib/auth.ts
 COPY lib/database/db_connect.ts ./lib/database/db_connect.ts
 COPY lib/model/color.ts ./lib/model/color.ts
+COPY lib/database/user.ts ./lib/database/user.ts
+COPY lib/database/db_connect.ts ./lib/database/db_connect.ts
```
radiantBear: This line is included twice

```suggestion
```

Dockerfile.deploy
```diff
@@ -2,17 +2,20 @@ FROM node:lts-alpine
 WORKDIR /app
 
 COPY package.json package-lock.json ./
+COPY tsconfig.json prisma.config.ts ./
+COPY prisma ./prisma
+
```
radiantBear: Please move this back down to avoid cluttering `git blame`

lib/auth.ts
```diff
@@ -92,7 +91,20 @@ export const auth = betterAuth({
...
+      beforeDelete: async user => {
+        /** Ideally this check would be transactionized with account deletion to
+         * prevent race conditions orphaning a list, but that does not appear
+         * possible with BetterAuth
+         */
+        const isLastAdmin = await getIsLastAdmin(user.id);
+
+        if (isLastAdmin) {
+          throw new APIError('BAD_REQUEST', {
+            message: 'Cannot delete a lists only admin account'
```
radiantBear:
```suggestion
            message: 'You are the only admin on at least 1 list with other users. You must promote someone to admin before deleting your account.'
```

cypress/e2e/auth/deleteAccount.cy.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    cy.findByLabelText('Confirm Delete Account').click();
+
+    cy.get('[data-toast="true"]')
+      .should('be.visible')
+      .within(() => {
+        cy.contains('Session expired.').should('be.visible');
+      });
+  });
+  it('Deletion succeeds with correct password', () => {
```
radiantBear, line 71:
```suggestion
  });

  it('Deletion succeeds with correct password', () => {
```

lilyvesstible (dismissed):

radiantBear (changes requested):
One more test that should be added is an e2e test validating that account deletion fails if the account is the last admin on a list.

lib/database/user.ts
```diff
@@ -165,3 +165,29 @@ export async function getRoleByList(
 
   return result ?? false;
 }
+
+/**
+ * Determines if a user is the last remaining admin in a list. For use to
+ * ensure when deleting a user, no list will be orphaned.
+ * @param userId
+ */
+export async function getIsLastAdmin(userId: string): Promise<boolean> {
```
radiantBear: I think this function name should be made more precise - it's not just checking for last admin, but also that there's at least 1 other user on the list.

Maybe `getIsOnlyAdminOnSharedList` 😬?

app/profile/components/AuthSettings.tsx
```diff
@@ -0,0 +1,306 @@
...
+          <div className='flex flex-col'>
+            <span className='text-base font-medium'>{method.title}</span>
+            <span className='text-sm text-gray-500'>{method.description}</span>
+          </div>
+        </div>
+        <Button
+          aria-label={method.actionLabel}
+          color={method.isCriticalAction ? 'danger' : 'default'}
+          size='sm'
+          variant='solid'
```
radiantBear: For consistency across the UI, we should use one of widely-used variants for these buttons
```suggestion
          variant='ghost'
```

lib/auth.ts
Ryangman: This error message is much longer than the Toasts have room to display. I agree that it's a nicer description but people will only see the full details if they inspect the actual http response.

lilyvesstible (dismissed):

lib/auth.ts
radiantBear: Bummer, toasts aren't smart enough to wrap to a second line? I think we need some clearer way of communicating why the failure happened; maybe you could add that message to the modal instead?

radiantBear (dismissed):
Thank you Ryan!

lilyvesstible (dismissed):

radiantBear (approved):

lilyvesstible (approved):

PR #123, opened 2026-03-15
fix: sign in button not replaced with profile icon on signin

radiantBear:
When a user signs in, the "Sign In" button is not replaced with their profile icon until they hard-refresh the page. This happens correctly on sign up, but not when signing back in, because of a typo in the asserted type of the sign in `onSuccess` handler.

## Testing Coverage
- Added an end-to-end test for sign in
- Added assertion to sign in & sign up tests that profile actions button (and profile icon) is displayed

## Reviewer Notes

It makes me a bit uncomfortable that this silently failed without a type check error or anything. Is there anything we can do besides e2e tests to catch this type of bug from BetterAuth's interface in the future?


Ryangman (dismissed):
These changes look good,. I see the deepsource alert is giving a D on code coverage, without any specific areas. Is this because you ignored coverage in vitest that cypress covered?

A-man37 (dismissed):

Ryangman (approved):
lgtm

A-man37 (approved):
🥳

PR #125, opened 2026-03-15
feat: allow members to be removed from a list

### LINKED ISSUE #57: Improve permissions system for multi-user lists

### LINKED ISSUE #58: Allow users to be removed from a list

radiantBear:
> [!IMPORTANT]
> Should not be merged before #127, which properly adds `<ConfirmModal />`

Completes the list permissions system by allowing users to be removed from lists they've been added to, as long as they're not the only admin remaining on the list.

image
image

## Related Issues

- Closes #58 
- Closes #57 

## Changes

- [x] Added API route for removing a member
- [x] Added database function for removing a member
- [x] Added GUI button for removing a member
- [x] Refactored `<MemberSettings />` to decrease JSX depth

### Tests added

- Full range of behavior from new `DELETE` API route
- Integration test validating members are removed client-side
- Added tests for coverage gaps in `<MemberSettings />`

### Tests fixed

- Added tests for proper RBAC on affected API routes
- Fixed wording for component label changes

### Tests needed
- End-to-end test validating full flow, including that the database call fails for the last admin on a list


components/ListSettings/MemberSettings.tsx
```diff
@@ -142,51 +231,19 @@ export default function MemberSettings({
...
-                  </SelectItem>
-                ))}
-            </Select>
-          </div>
-        ))
-        .toArray()}
-    </span>
+      </div>
+      <ConfirmModal
+        description='This action will unassign them from all tasks and revoke their access. Cannot be done if this would leave the list without an admin.'
```
Ryangman:
```suggestion
        description='This action will unassign them from all tasks and revoke their access. This cannot be done if it would leave the list without an admin.'
```

components/ListSettings/__tests__/MemberSettings.test.tsx
```diff
@@ -188,6 +188,38 @@ describe('Adding members', () => {
     });
     expect(onMemberEvent).not.toHaveBeenCalled();
   });
+
+  it('Does not allow new member role to be empty', async () => {
```
Ryangman: This name doesn't seem to accurately represent what is tested here. This name reads like a backend test that when adding a new member it will always give them a non-empty role. 
```suggestion
  it('Role picker renders the name of member role selected from role dropdown', async () => {
```

components/ListSettings/MemberSettings.tsx
```diff
@@ -102,35 +179,47 @@ export default function MemberSettings({
...
           type: 'UpdateMemberPermissions',
-          id: userId,
+          id: member.user.id,
           role: role.id
         });
       })
       .catch(addToastForError);
   }
 
+  function handleRemoveMember() {
```
Ryangman, line 202: When you remove yourself from a list, the list settings modal remains open, and the list visible and operational until the page is refreshed. This shouldn't pose a security issues as all routes verify a user's authorization, but this is a clunky experience for a user, and could be a security fault if any route is improperly secured.

Ryangman (changes requested):
These changes look good, there are a couple naming related improvements that could be made, but the main issue is when removing yourself from a list you still have view access until you refresh the page. It's probably not worth forcing a page refresh but if there is a simple way to clean up the user experience there it would be worthwhile.

Ryangman (dismissed):
This is looking better! The modal behavior is now as I would imagine a user should expect. One minor thing, when deleting yourself from a list, the `router.replace()` sends you back to your main list page (as expected), but the list name persists in the sidebar until reload. I would understand if this isn't worth fixing at this point though.

lib/database/list.ts
```diff
@@ -379,3 +379,51 @@ export async function deleteTag(id: string): Promise<boolean> {
...
+ * @param listId The list to remove a member from
+ * @param userId The user to revoke access for
+ * @returns Whether the database accepted the update
+ */
+export async function deleteListMember(
+  listId: string,
+  userId: string
+): Promise<boolean> {
+  try {
+    await prisma.$transaction(
```
sabri414, line 395: This is looking good overall, and the member-removal flow is much closer. I’m holding approval for one backend issue before the merge:
deleteListMember() currently deletes the ListMember row inside the transaction, but clears ItemAssignee rows afterward. If that second step fails, the API returns a 500 even though the member was already removed, which leaves the operation in a partial-success state and makes retries unsafe.
I’d like to see the member removal and assignee cleanup happen atomically in the same transaction.

sabri414 (changes requested):

sabri414:
I see the transactional fix in commit 5b47ce7c..., but PR #125 still appears to be on 02e4de35..., where itemAssignee.deleteMany() is outside the transaction. Once that fix is actually on this PR branch, I’m happy to re-review.

sabri414 (dismissed):
The latest relationship change addresses my main concern with the member-removal flow by making the assignee cleanup part of the database relationship rather than a separate step after the transaction.

Good job!

radiantBear:
Thanks for the quick reviews, Sabri! I actually wasn't done with my changes yet.

While you're right that it's theoretically possible to have a partial-success during deletion that could cause inconsistent state, I'm not aware of any condition that would cause an error to be thrown when deleting from `ItemAssignee`. A bigger concern to me is that the `READ COMMITTED` isolation level (necessary to prevent race conditions while checking if the user is the last admin on a list) could open up a race condition where an `ItemAssignee` is created during deletion and therefore isn't included in the `DELETE`.

However, unanticipated errors are possible and you're right to point out that we should ensure that database operations are fully atomic. I think the solution is to update the foreign key constraints so that `ItemAssignee` has a M:1 relationship with `ListMember`, not `User`, then rely on `ON DELETE CASCADE` to remove the relevant `ItemAssignee` rows when a member is deleted.

I just added 2 commits that provide ways to accomplish this. The first is simpler, but I don't completely love it since the application layer must ensure that the `listId` in `ItemAssignee` matches the `ItemAssignee.itemId -> Item.sectionId -> Section.listId`. The second fixes this, but adds a fair bit of complexity to the foreign key definitions. Rather than either of these, it would probably be simpler to give `ListMember` a unique `id` primary key, but making this a `@db.Char(16)` like all the other `id`s would make migration fairly difficult, so I didn't pursue it.

I would appreciate feedback on which approach to move forward with and another pair of eyes checking this change to ensure that this race condition is prevented and the table relations are as idiomatic as possible (from @Ryangman too).

sabri414 (commented):
Thanks for the detailed write-up. I’d move forward with the second approach (Strengthen constraints).

My reasoning is that Fix relationship addresses the atomic delete concern, but it still relies on the application to keep ItemAssignee.listId consistent with the list implied by itemId. The stronger-constraints version pushes that invariant into the database by tying:

ItemAssignee (userId, listId) to ListMember
ItemAssignee (itemId, listId) to Item
Item (sectionId, listId) to ListSection

That feels like the safer and more idiomatic direction to me, even if it adds some schema complexity, because it closes the race/consistency gap instead of just moving it around.

I took another pass, and I’m comfortable that this version fixes the original partial-success concern in a more robust way.

sabri414 (approved):

Ryangman (approved):
I'd personally prefer the second option. Relying on the CASCADE operations feels safer than assuming application layer checks are done correctly/consistently. 

This seems functional from my testing (hence my approval), but an e2e test would be useful for this behavior.

PR #126, opened 2026-03-15
fix: member permission updates rejected by database

### LINKED ISSUE #122: Member role change fails with no error logged

radiantBear:
Fixes a bug preventing list members' roles from being changed.

## Related Issues
- Closes #122 

## Testing Coverage

This bug is only detectable via end-to-end tests. This behavior should be verified when those test suites are added.


Ryangman (approved):
lgtm

sabri414 (approved):
Looks good to me!

PR #127, opened 2026-03-15
style: use custom confirm modal instead of browser default

radiantBear:
Instead of using the browser's default `confirm` dialog, this uses a custom confirm component that matches Tasktix's style.

image
image
image

## Changes

- [x] Added confirm modal
- [x] Switched default `confirm` calls to use modal instead

### Tests added

- Custom modal displays all fields
- Confirm and cancel buttons work

### Tests updated

- Tests that mocked `confirm()` now press the confirm/cancel button in the custom modal

## Reviewer Notes

How does this component's style look? Any changes we should make now?

HeroUI v3 [adds a component for this purpose](https://v3.heroui.com/docs/react/components/alert-dialog) so this is to some degree a stopgap.


lilyvesstible (dismissed):

lilyvesstible (approved):

sabri414 (approved):
Looks good to me overall. I like the move away from the browser default confirm dialog to a reusable modal that matches the app style, and the updated tests cover the new flow well.
One small nit: in `ConfirmModal`, the description uses `text-gray-500`. Since this component will show in both themes, I think a semantic token like `text-default-500` might fit better and stay more consistent with HeroUI theme styling.
Other than that, I’m good with this.

PR #128, opened 2026-03-16
fix: better-auth session update conflict

### LINKED ISSUE #121: BetterAuth's session update intermittently fails when loading the homepage
Server error log:
```
2026-03-14T20:54:02.254Z ERROR [Better Auth]: INTERNAL_SERVER_ERROR Error [PrismaClientUnknownRequestError]: 
Invalid `prisma.session.update()` invocation:


Error occurred during query execution:
ConnectorError(ConnectorError { user_facing_error: None, kind: QueryError(Server(MysqlError { code: 1020, message: "Record has changed since last read in table 'Session'", state: "HY000" })), transient: false })
    at async Object.update (.next/server/chunks/ssr/_823f051b._.js:438:9450)
    at async updateWithHooks (.next/server/chunks/ssr/_823f051b._.js:44:7284)
    at async Object.updateSession (.next/server/chunks/ssr/_823f051b._.js:44:13701)
    at async (.next/server/chunks/ssr/_823f051b._.js:44:29446)
    at async g (.next/server/chunks/ssr/_823f051b._.js:41:195217) {
  clientVersion: '6.19.2'
}
⨯ [Error [APIError]: Failed to get session] {
  status: 'INTERNAL_SERVER_ERROR',
  body: [Object],
  headers: {},
  statusCode: 500,
  digest: '3914836991'
}
```

radiantBear:
> [!IMPORTANT]
> 
> Likely changing approach; see https://github.com/Tasktix/Tasktix/issues/121#issuecomment-4065842778. This would be good for allowing the homepage's `redirect` to happen server-side instead of relying on client-side redirects that seem to occasionally fail as well.

When loading the Tasktix homepage, users are intermittently met with a message that a server-side error occurred. This (hopefully) resolves that issue.

## Related Issues

- Closes #121 

## Changes

- [x] Removed duplicate Prisma client instantiation
- [x] Changed `getUser` calls to `useAuth` where possible

## Testing Coverage

This bug seems to be triggered by a lost race condition, so I have been unsuccessful in consistently reproducing this error. While I have never encountered it with these fixes, we will need to monitor logs in prod to ensure the issue is truly resolved.

## Reviewer Notes

The error message for this bug is not very helpful, especially since it doesn't contain a stack trace. My guess is that the cause is BetterAuth regenerating a session user's session multiple times simultaneously, which causes a race condition. If one session regeneration starts before the other but finishes after, this causes the error to be thrown.

I made 2 attempts to resolve this:
- BetterAuth had a separate Prisma client than the rest of the app. This hopefully isn't the root cause (or horizontal scaling will be impossible), but is worth fixing.
- This error has only been encountered on the homepage. `app/layout.tsx` makes one call to `getUser` and `app/page.tsx` makes a second, which could be the race condition trigger. `app/page.tsx` has been updated to use the Context object provided by `app/layout.tsx` instead of fetching the user directly to attempt to resolve this.

### DIFF
--- a/Dockerfile.deploy
+++ b/Dockerfile.deploy
@@ -2,16 +2,18 @@ FROM node:lts-alpine
 WORKDIR /app
 
 COPY package.json package-lock.json ./
-RUN npm ci --ignore-scripts
+COPY tsconfig.json prisma.config.ts ./
+COPY prisma ./prisma
+
+RUN npm ci --ignore-scripts && npm run prisma:generate
 
 # Stub import of server-only, this interprets test environment as client component
 # server-only still prevents server code being run on client in development/production
 RUN echo "" > "/app/node_modules/server-only/index.js"
 
 COPY scripts/test_service_entry.sh ./entry.sh
-COPY tsconfig.json prisma.config.ts ./
-COPY prisma ./prisma
 COPY lib/auth.ts ./lib/auth.ts
+COPY lib/database/db_connect.ts ./lib/database/db_connect.ts
 COPY lib/model/color.ts ./lib/model/color.ts
 
 CMD [ "sh", "entry.sh" ]
--- a/app/page.test.ts
+++ b/app/page.test.ts
@@ -19,25 +19,20 @@
 import { redirect } from 'next/navigation';
 
 import Home from '@/app/page';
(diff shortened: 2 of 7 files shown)


radiantBear:
Superseded by #130

PR #134, opened 2026-04-08
feat: add backend to support section change 

Ryangman:
Modifies the `PATCH /api/item/[id]` route to support section changes if the body includes `{'sectionId', [sectionId]}`

Related to #76 
>[!Note]
>Current CI results from hash `d6ab359f4e478af14dd50c70ea165da775fdf48a` are due to:
>- Test coverage not in scope for this PR
>- Linting Issues from the `change-section-dropdown` branch

### DIFF
--- a/app/api/item/[id]/route.ts
+++ b/app/api/item/[id]/route.ts
@@ -21,6 +21,7 @@ import { getRoleByItem } from '@/lib/database/user';
 import {
   deleteListItem,
   getListItemById,
+  updateItemSection,
   updateListItem
 } from '@/lib/database/listItem';
 import { getUser } from '@/lib/session';
@@ -28,9 +29,8 @@ import { ZodListItem } from '@/lib/model/listItem';
 
 const PatchBody = ZodListItem.omit({
   id: true,
-  sectionIndex: true,
-  sectionId: true
-}).partial();
+  sectionIndex: true
+}).partial()
 
 export async function PATCH(
   request: Request,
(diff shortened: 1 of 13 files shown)


lilyvesstible (approved):

lib/database/listItem.ts
```diff
@@ -193,3 +206,51 @@ export async function unlinkAssignee(
...
+ * @param sectionId The sectionId of the target Section
+ */
+export async function updateItemSection(
+  item: ListItem,
+  sectionId: string
+): Promise<boolean> {
+  try {
+    await prisma.$transaction(
+      async tx => {
+        const targetSectionCount = await tx.item.count({
```
radiantBear:
```suggestion
        const targetSectionItems = await tx.item.count({
```

lib/database/listItem.ts
```diff
@@ -193,3 +206,51 @@ export async function unlinkAssignee(
...
+          where: { sectionId }
+        });
+        const originalIndex = item.sectionIndex;
+
+        await tx.item.update({
+          where: { id: item.id },
+          data: {
+            sectionId,
+            sectionIndex: targetSectionCount + 1
```
radiantBear:
```suggestion
            sectionIndex: targetSectionCount
```

app/api/item/[id]/route.ts
```diff
@@ -58,25 +59,37 @@ export async function PATCH(
...
-      : null;
-  if (requestBody.dateCompleted !== undefined)
-    item.dateCompleted = requestBody.dateCompleted
-      ? new Date(requestBody.dateCompleted)
-      : null;
-
-  const result = await updateListItem(item);
-
-  if (!result) return ServerError.Internal('Could not update item');
+  if (requestBody.sectionId !== undefined) {
```
Comment (radiantBear) [inline app/api/item/[id]/route.ts]: I think it would be good to align with SonarQube's recommendation here

app/api/item/[id]/route.ts
Comment (radiantBear) [inline app/api/item/[id]/route.ts:53]: Is it really reasonable to use the `canUpdateItems` permission for changing a list item's section? That feels more like a section-level or even list-level modification to me, at least as far as expected permissions go.

radiantBear (commented):
> [!CAUTION]
> Off by one error!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!?????????????????????????????????????
> 
> Double-check as part of #76

app/api/item/[id]/route.ts
Comment (Ryangman) [reply app/api/item/[id]/route.ts:53]: 1) We don't have section level permissions. 2) `canUpdateList` I interpreted as list metadata changes (e.g. removing time tracking, using due dates, etc. ). 

I may be seeing this differently the use case I imagined for this changing section is mostly for kanban boards, and moving things from backlog->open->ready for review, etc. I feel like that should be an action that a the user who is allowed to update the specific item (e.g time estimates, tags) should probably have the authority to do, and not need to request a list owner to move the item.

Comment (radiantBear) [reply app/api/item/[id]/route.ts:53]: Good point, especially if we add a feature that automatically changes sections when an item's status changes. It still feels a bit wrong to me since it has visual implications beyond the single item, but that's probably the best permission requirement for now.

radiantBear (approved):

PR #115, opened 2026-03-12
refactor: simplify data structure for List component

radiantBear:
Simplifies the `<List />` component's data structure to remove duplicated data (e.g. a global list of `Tag` objects plus full `Tag` objects on each item they're linked to; a global list of `Member` objects plus full `Member` objects on each item they're linked to) and nested state. This follows [React's best practices](https://react.dev/learn/choosing-the-state-structure#avoid-deeply-nested-state) for normalizing complex state to avoid deep nesting that is unperformant to update. This also improves the independence of state updates to increase the safety of API calls with SSE side effects (see the `AddNewTag` bullet points in the changes section).

## Related Issues

- Unblocks #109
- Unblocks #125 
- Needs #100 to be merged to normalize its data, too
- Structuring changes will affect #76

## Changes
- [x] Fixes callbacks to use 'on' prefix
- [x] Moves nested data to top-level maps to deduplicate data & simplify state updates
- [x] Fully separate `AddNewTag` state update into separate state updates
  - Previously created full new tag for list item, in case the global list of tags hadn't yet updated to include the new one
  - Now, new tags just have their ID added to the list item - same as any other tag
  - If this happens before the global list of tags updates, upon the global list update, the tag will automatically appear on the list item
  - This means it's safe to perform 2 separate API calls with SSE side effects to immediately link a new tag to an item upon creation
- [x] Removes already-broken today view instead of showing broken implementation

### Tests Added
- Tests for list metadata changes via SSE
- Tests for updating list membership
- Tests for updating tags (names & colors)
- Tests for adding/deleting sections

### Tests Fixed
- Fixed tests for updated type signatures

### Tests Needed
- Tests for reordering items not included (likely to be changed by #47)
- Tests for assignees not included: more future work needed on assignees; not in scope of list refactor

PR #146, opened 2026-04-09
fix: prevent non-shared lists from blocking account deletion

### LINKED ISSUE #145: account deletion fails for non-shared lists
Database checks that there is no other admin, should also ensure that there is one other member

Ryangman:
A bug in the database interface to check if a user was the last admin of a list proactively prevented deletion of users who were the only admin of a list that wasn't shared.

This fixes that bug and adds e2e testing to prevent future issues. 

## Related Issues

Closes #145 

## Changes
- [x] fixed bug in database function that checks last admin
- [x] added e2e test to verify behavior
- [x] added cypress command to create dummy test list  

## Testing Coverage
Tested via e2e tests, database action cannot be verified by in vitest

Test coverage issues from deepsource are out of scope for this PR, and will mostly be resolved with the normal vs Github List Modal

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -73,11 +73,22 @@ Cypress.Commands.add('login', (username: string, password: string) => {
     .and('not.be.null');
 });
 
+Cypress.Commands.add('createList', (listName: string) => {
+  cy.visit('/list');
+
+  cy.get('button .bi-plus').click();
+  cy.findByPlaceholderText('List name').type(listName);
+  cy.get('button[type="submit"]').click();
+
+  cy.location('pathname').should('match', /^\/list\/[a-zA-Z0-9]{16}$/);
+});
+
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace Cypress {
     interface Chainable {
       login(email: string, password: string): Chainable<void>;
+      createList(listName: string): Chainable<void>;
     }
   }
 }
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -76,9 +76,9 @@ Cypress.Commands.add('login', (username: string, password: string) => {
 Cypress.Commands.add('createList', (listName: string) => {
   cy.visit('/list');
 
-  cy.get('button .bi-plus').click();
+  cy.findByLabelText('Create new list').click();
   cy.findByPlaceholderText('List name').type(listName);
-  cy.get('button[type="submit"]').click();
+  cy.findByLabelText('Submit list').click();
 
   cy.location('pathname').should('match', /^\/list\/[a-zA-Z0-9]{16}$/);
 });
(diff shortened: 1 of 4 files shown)


cypress/support/commands.ts
```diff
@@ -73,11 +73,22 @@ Cypress.Commands.add('login', (username: string, password: string) => {
     .and('not.be.null');
 });
 
+Cypress.Commands.add('createList', (listName: string) => {
+  cy.visit('/list');
+
+  cy.get('button .bi-plus').click();
```
radiantBear: It would be better to use an aria label or name attribute to find this button rather than a CSS class. If it doesn't have one yet, could you please add that?

cypress/support/commands.ts
```diff
@@ -73,11 +73,22 @@ Cypress.Commands.add('login', (username: string, password: string) => {
...
+  cy.findByPlaceholderText('List name').type(listName);
+  cy.get('button[type="submit"]').click();
+
+  cy.location('pathname').should('match', /^\/list\/[a-zA-Z0-9]{16}$/);
+});
+
```
radiantBear, line 85: More generally, is this a helper function we should even define? It seems to me that it might be better to just expand the seed script to create a few different variations of lists for testing.

radiantBear (changes requested):
Thanks Ryan! That logic tweak makes sense; easy to accidentally overlook that edge case. Thanks for adding an e2e test for it!

cypress/support/commands.ts
Ryangman: I think for these end to end tests we should be using the users flow through the UI as much as possible, for example with my original review on #115, a user couldn't create a list because what the api sent the to the database interface and what the database expected where out of sync. If we manually seeded a list, even e2e tests wouldn't have caught that, but this would.

radiantBear: Good point! On the other hand, [Cypress' best practices](https://docs.cypress.io/app/core-concepts/best-practices#Organizing-Tests-Logging-In-Controlling-State) note that "taking shortcuts" and "programmatically taking control of application state" are good things. Especially when e2e testing more complex features, like filtering, it seems better to use DB seeding and have a separate test to validate that list creation works.

I think this is fine for now, but we should be careful about how much test-state generation we do by interacting with the site.

Ryangman: Long term i'd agree, but until we have a test designed to verify list creation, I don't thing we should assume it works by seeding. Id say our best practice would be as soon as we have a test that thoroughly verifies something works we can seed that behavior in other tests, but until that test exists use the user flow. Kind of like how sign up is now.

radiantBear (dismissed):

sabri414 (dismissed):
Looks good!

radiantBear (approved):

A-man37 (approved):

sabri414 (approved):

PR #149, opened 2026-04-27
feat: replace List creation dialog with modal

### LINKED ISSUE #26: List creation dialog persists after creation
When creating a new list, the dialog does not automatically clear and disappear. Instead, the user must click out of it to dismiss the creation dialog.

https://github.com/user-attachments/assets/b7695165-3a3c-4e80-a804-5cc8befd1de9

### LINKED ISSUE #129: Create modal for picking GitHub vs regular list

Ryangman:
Replaces list creation text input with a modal that allows users to optionally link a list to a GitHub Repository if they have configured OAuth with GitHub and provided access to at least one repository.

## Related Issues

Closes #129
Closes #26 
Related to #94 

## Changes
- [x] Replace List Creation Dialog with Modal
- [x] Add Database function to determine if user is linked to GitHub from server
- [x] Add API route `/user/[id]/github` to fetch repositories that a user has made accessible to the App
- [x] Add migration to database to track the `repoId` a list is tracking
- [x] Updated Documentation for how to configure Tasktix GitHub App
- [x] Add Tests

## Testing Coverage
Added tests for API that were created/modified. Did not test Sidebar as Sabri's #99 will add tests for that. Github Integration directory is not reasonable to test as it is exclusively network interface with Github, and even e2e test would require creating a fake Github account with fake repositories.


lib/database/user.ts
```diff
@@ -194,3 +194,23 @@ export async function getIsOnlyAdminOnSharedList(
 
   return Boolean(orphanedList);
 }
+
+/**
+ * Determines whether a user has an account linked to github. This is necessary
+ * as BetterAuth does not provide a clean way to determine this from the server.
+ * (The only solution i've found is to try and fetch the access token, but it
+ * fails loudly)
```
radiantBear: When you say it "fails loudly," does that just mean it throws an error, or does it call `exit()`? If it's just an error that could be wrapped in `try {} catch {}`, I think that would be preferable over the latency & unnecessary load created by a database query. If you're strongly not in favor of that style, you could add a `tryOrUndefined` helper to `lib/util.ts` so that you can use the errors-as-values style here.

eslint.config.mjs
```diff
@@ -116,7 +116,11 @@ export default defineConfig([
     }
   },
   {
-    files: ['app/api/**/*.ts', 'lib/database/**/*.ts'],
+    files: [
+      'app/api/**/*.ts',
+      'lib/database/**/*.ts',
+      'lib/integration/**/*.ts'
```
radiantBear, line 122: If we're gonna allow `console` in `lib/integration`, I think it'd be good to include `import 'server-only'` in those files

lib/integration/github/account.ts
```diff
@@ -0,0 +1,71 @@
...
+
+type ListRepositoriesResponse =
+  Endpoints['GET /user/installations/{installation_id}/repositories']['response']['data'];
+
+/**
+ * Fetches the repositories made available to a Github App installation by a
+ * User Access Token. This function returns the full object provided by the
+ * Github REST API, which should often be filtered as it contains far more
+ * information than needed for any Tasktix use case.
+ * @param accessToken
```
radiantBear: Please don't forget to provide a description for this parameter!

lib/integration/github/account.ts
```diff
@@ -0,0 +1,71 @@
...
+export async function getAccessibleRepositories(
+  accessToken: string
+): Promise<ListRepositoriesResponse['repositories'] | false> {
+  const octokit = new Octokit({ auth: accessToken });
+
+  try {
+    const installations = await octokit.request('GET /user/installations', {
+      headers: {
+        'X-GitHub-Api-Version': '2026-03-10'
+      }
```
radiantBear, line 40: I'm a bit surprised this is done manually. Does `octokit` really not manage the GitHub API version based on which version of the SDK is in use?

lib/integration/github/account.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Fetches the repositories made available to a Github App installation by a
+ * User Access Token. This function returns the full object provided by the
+ * Github REST API, which should often be filtered as it contains far more
+ * information than needed for any Tasktix use case.
+ * @param accessToken
+ */
+export async function getAccessibleRepositories(
+  accessToken: string
+): Promise<ListRepositoriesResponse['repositories'] | false> {
```
radiantBear: Is there a particular reason `SimplifiedRepo` isn't the return type here? Where possible, I think wrappers around third-party code should use our own interfaces to decrease coupling in case we need to separate from that library/SDK in the future. Also, thinking about more general integrations, we'll likely want to allow similar list <-> third-party linking for other tools (using a name and description), so maybe the type should be even more general than `SimplifiedRepo`.

lib/model/list.ts
```diff
@@ -31,7 +31,8 @@ export const ZodList = z.strictObject({
   color: z.enum(namedColors),
   hasTimeTracking: z.boolean(),
   hasDueDates: z.boolean(),
-  isAutoOrdered: z.boolean()
+  isAutoOrdered: z.boolean(),
+  repoId: z.int()
```
radiantBear: If `repoId` is optional in the model, it should probably be optional by default in the Zod schema

app/api/user/[id]/github/route.ts
```diff
@@ -0,0 +1,70 @@
...
+import { ClientError, ServerError, Success } from '@/lib/Response';
+import { getUser } from '@/lib/session';
+import { getIsAccountLinkedToGithub } from '@/lib/database/user';
+
+export const dynamic = 'force-dynamic'; // defaults to auto
+
+export async function GET(
+  request: Request,
+  { params }: { params: Promise<{ id: string }> }
+) {
```
Comment (radiantBear) [inline app/api/user/[id]/github/route.ts:34]: Please don't forget to add a docblock comment here!

app/api/user/[id]/github/route.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  const user = await getUser();
+
+  if (!user) return ClientError.Unauthenticated('Not logged in');
+
+  if (user.id !== id) return ClientError.Forbidden('Insufficient permissions');
+
+  const isLinked = await getIsAccountLinkedToGithub(user.id);
+
+  if (!isLinked) return ClientError.BadRequest('Account not linked to Github');
```
Comment (radiantBear) [inline app/api/user/[id]/github/route.ts]: I think `NotFound` would be better semantics than `BadRequest` here - there was nothing wrong with the request contents; there just isn't any data to return.

app/api/user/[id]/github/route.ts
```diff
@@ -0,0 +1,70 @@
...
+    accessToken = response.accessToken;
+  } catch (error) {
+    console.error('BetterAuth API Error:', error);
+
+    return ServerError.Internal('Failed to fetch valid Access Token');
+  }
+  const repositories = await getAccessibleRepositories(accessToken);
+
+  if (!repositories)
+    return ServerError.Internal('Failed to fetch accessible repositories');
```
Comment (radiantBear) [inline app/api/user/[id]/github/route.ts]: I wonder if `NotFound` would be more appropriate in this case, too - if the user just hasn't created any repos yet, that isn't a Tasktix server error.

components/Sidebar/Sidebar.tsx
```diff
@@ -135,62 +149,176 @@ export function NavItem({
   );
 }
 
-function AddList({ addList }: { addList: () => unknown }) {
+function AddList({ openListDialog }: { openListDialog: () => unknown }) {
```
radiantBear: Following React conventions for event handler callbacks, this should start with `on`:
```suggestion
function AddList({ onCreate }: { onCreate: () => unknown }) {
```
(or maybe `onRequestCreate`, `onPressNew`, etc...)

components/Sidebar/Sidebar.tsx
```diff
@@ -135,62 +149,176 @@ export function NavItem({
...
     >
       <Plus size={'1.25em'} />
     </Button>
   );
 }
 
-function NewItem({
-  finalize,
-  remove
+/**
```
radiantBear: Since this is a pretty big chunk that includes its own subcomponents (e.g. GitHub), I think it would make sense to move this to a new file.

Good for hopefully reducing merge conflicts with Sabri, too \:)

components/Sidebar/Sidebar.tsx
```diff
@@ -135,62 +149,176 @@ export function NavItem({
...
+ * name and optionally link it to a specified github repository (if configured)
+ * @param isOpen The `isOpen` parameter from HeroUI's `useDisclosure`
+ * @param onOpenChange The `onOpenChange` parameter from HeroUI's `useDisclosure`
+ * @param submitList A handler function used to submit the list to the API
+ * @param isGithubConfigured Bool representing if github integration is configured
+ * @returns
+ */
+function CreateListModal({
+  isOpen,
+  submitList,
```
radiantBear: Event handler callbacks should start with `on`
```suggestion
  onSubmit,
```

components/Sidebar/Sidebar.tsx
```diff
@@ -135,62 +149,176 @@ export function NavItem({
...
+                    Cancel
+                  </Button>
+                  <Button
+                    aria-label='Submit list'
+                    color='primary'
+                    type='submit'
+                  >
+                    Confirm
+                  </Button>
+                </div>
```
radiantBear: For consistent styling, this probably makes more sense as the `<ModalFooter>`. You could wrap `<ModalBody>` and `<ModalFooter>` in the `<Form>` to make this work

components/Sidebar/Sidebar.tsx
```diff
@@ -135,62 +149,176 @@ export function NavItem({
...
+          // JSX depth already abstracted to reasonable level skipcq: JS-0415
+          <>
+            <ModalHeader className='flex gap-2 items-center'>
+              Create New List
+            </ModalHeader>
+            <ModalBody className='text-sm text-default-500'>
+              <Form onSubmit={e => handleSubmitList(e, onClose)}>
+                <Input
+                  color='primary'
+                  placeholder='List name'
```
radiantBear: For consistency across the UI. `label` is generally used over `placeholder`
```suggestion
                  label='List name'
```

components/Sidebar/Sidebar.tsx
```diff
@@ -63,31 +86,22 @@ export default function Sidebar({ lists }: { lists: List[] }) {
...
           .sort((a, b) => (a.name > b.name ? 1 : 0))
           .map(list => (
             <NavItem key={list.id} link={`/list/${list.id}`} name={list.name} />
           ))}
-        {addingList ? (
-          <NewItem finalize={finalizeNew} remove={removeNew} />
-        ) : null}
       </NavSection>
+      <CreateListModal
+        isGithubConfigured={oauthConfig.githubEnabled}
```
radiantBear: I think it would be slightly cleaner to let `<CreateListModal>` determine for itself whether GitHub integration is enabled (unless I'm missing another place `oauthConfig` is needed

components/Sidebar/Sidebar.tsx
```diff
@@ -135,62 +149,176 @@ export function NavItem({
...
+        >
+          {availableRepos.map(repo => (
+            <SelectItem key={repo.id} description={repo.description}>
+              {repo.name}
+            </SelectItem>
+          ))}
+        </Select>
+        <p className='py-4'>
+          To access more repositories{' '}
+          <Link isExternal showAnchorIcon href='asdf' underline='hover'>
```
radiantBear: Should this link's `href` be populated? If not, it may be better to leave this `<p>` out until a later PR

radiantBear (changes requested):
Thank you Ryan, this is a great start for integrating GitHub! I left some thoughts on the implementation below; I'll come back to review the tests soon.

eslint.config.mjs
Ryangman: server only is imported in those files

lib/integration/github/account.ts
Ryangman: Not here, because we are using raw request object from Octokit. For things like creating Issues, PRs, etc. they have built in methods that handle all API logic, but for these account routes they do not have custom methods we have to use the raw request.

app/api/user/[id]/github/route.ts
Comment (Ryangman) [reply app/api/user/[id]/github/route.ts]: The two situations this error would be triggered are:

1) The user has made no repositories accessible to the installation
2) The Github API is down, or otherwise unresponsive

I think the latter is a much more likely occurrence, and when the first issue makes sense as NotFound,  but it is abundantly clear in the modal that if you don't see repositories, visit and configure your app installation. While the github issue is technically is an *external* server error, I'm don't think NotFound makes sense in that situation.

components/Sidebar/CreateListModal.tsx
```diff
@@ -0,0 +1,219 @@
...
+              {repo.name}
+            </SelectItem>
+          ))}
+        </Select>
+        <p className='py-4'>
+          To access more repositories{' '}
+          <Link
+            isExternal
+            showAnchorIcon
+            href='https://github.com/apps/tasktix;'
```
radiantBear: Currently returns a 404
```suggestion
            href='https://github.com/apps/tasktix'
```

lib/integration/github/types.ts
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+
```
radiantBear, line 18: Maybe not super necessary since there's no logic here, but I think this would be good for consistency since we're disabling client-side lints for this whole directory.
```suggestion

import 'server-only';

```

app/api/list/route.test.ts
```diff
@@ -0,0 +1,105 @@
...
+        body: JSON.stringify({ color: 'Amber', name: 'list name' })
+      })
+    );
+
+    expect(response.status).toBe(201);
+    expect(createList).toHaveBeenCalledExactlyOnceWith(
+      expect.objectContaining({ color: 'Amber', name: 'list name' })
+    );
+  });
+
```
radiantBear, line 62: I think it would be good to have a regression test to validate that the `repoId` gets passed on to the database when provided. I know that's a pretty minor case, but it should be quick to add and will mean that the new feature is actually tested.

radiantBear (commented):
Thanks Ryan, this is looking really good! I've reviewed all the tests and the change you've made and just have a couple last thoughts.

radiantBear (approved):
Thanks Ryan! This looks good

A-man37 (approved):
Looks good!

PR #150, opened 2026-04-27
style: update /about with new features, etc.

radiantBear:
Updates the `/about` page in preparation for an initial demo.

image
image
image
image

## Changes

- [x] Updated styling to improve accessibility/readability
- [x] Updated feature list to highlight changes from the last 6 months
- [x] Added an auto-playing video demonstrating how Tasktix is used

## Testing Coverage

N/A - no logic changes, just informational changes

## Reviewer Notes

Let me know if you think any of the descriptions could be phrased better!


Ryangman (changes requested):
This looks a lot better!

Two Issues i've found:
- The video should have a a light mode variant, it looks rather jarring with light/dark contrast
- The footer doesn't appear to be rendering correctly. On my screen it renders slightly below the page on first load, but then is static after that

lilyvesstible (dismissed):

Ryangman (approved):
Approving this with for early merge with Project Landing Page Assignment due. Bugs with theme switcher and video improvements should be made sooner rather than later

PR #152, opened 2026-04-27
fix: theme switching for about page demo video

### LINKED ISSUE #151: About page video theme doesn't switch with theme switcher; requires hard reload

radiantBear:
Two-for-one fix that
- Fixes a typo keeping the light-mode video poster/preview image from loading, and
- Fixes a bug where the video doesn't change theme when the theme switcher is used, unless the page is reloaded

## Related Issues

- Closes #151 

## Testing Coverage

I don't think there's a way to add unit tests to detect this regression (since it has to do with React's behavior when `<video>` components' props are updated) and I don't think it's particularly worth end-to-end testing this specifically, so there are no new tests here.


Ryangman (approved):
lgtm

A-man37 (approved):
Seems simple enough. :)

PR #154, opened 2026-04-29
feat: allow disabling local sign-up in favor of SSO

### LINKED ISSUE #153: Add environment variable flag for disabling sign-up
Self-hosters (particularly using OIDC for authentication) may wish to disable sign-up so they can publicly expose the site without allowing arbitrary people on the Internet to use their service

radiantBear:
Allows self-hosters to disable local authentication if they'd like to solely rely on 3rd-party authentication. 

> [!NOTE]
> These images only show the top of the screen, so the aspect may look a bit strange for the messages that look like they're at the bottom. They're actually still in the top 1/3 of the screen.

image
<br>
image
<br>
image
<br>
image
<br>
image
<br>
image

## Related Issues

- Closes #153 

## Changes

- [x] Fixed minor inconsistency between "Sign in" and "Log out"
- [x] Added ability to shut off local sign-in using environment variables
  - Still theoretically possible to log *in* with an existing local account, but new accounts cannot be created
- [x] Fixed variable names to reflect broader scope
- [x] Simplified flow with only 1 auth provider to immediately redirect to their service

### Tests added

- Validate that local auth isn't shown when disabled
- Validate that a single configured auth provider results in an immediate redirect

### Tests modified

- Moved OAuth UI tests to sign-in page tests, where most of the logic is now tested

## Reviewer Notes

Styling for the 2-SSO-but-no-local page looks a bit funky. I'm open to suggestions to fix it, but I doubt people would tend to use it...


app/signIn/page.tsx
```diff
@@ -18,25 +18,87 @@
...
+      void handleOAuth('custom', setLoggedInUser, authConfig); // skipcq: JS-0098
+
+    return (
+      <main className='flex grow justify-center items-start mt-40 w-xl mx-auto'>
+        <Alert
+          color='success'
+          title='Redirecting for SSO sign in...'
+          variant='flat'
+        />
+      </main>
```
Ryangman, line 80: It'd probably be worth adding a button/link to manually redirect to SSO if the redirect fails. This will particularly be an issue if users have their browsers configured to block redirects.

.env.example
```diff
@@ -7,6 +7,8 @@ DB_PASSWORD=password
 
 BETTER_AUTH_SECRET=generate a random secret value, 32 char min
 BETTER_AUTH_URL=http://localhost:3000 # Base URL of APP
+
+DISABLE_LOCAL_AUTH=false
 ENABLE_PASSWORD_STRENGTH_CHECK=true
```
Ryangman: `DISABLE_LOCAL_AUTH` and `ENABLE_PASSWORD_STRENGTH_CHECK` should probably use the same semantic style/polarity. I'd probably prefer using `DISABLE_{FEATURE}` since these are both features that _default_ to being on

Ryangman (changes requested):
This seems like a useful feature, i haven't reviewed testing updates yet with cypress still failing, but i've got some minor changes for the feature itself

app/signIn/page.tsx
radiantBear: Good point! It doesn't really look like BetterAuth supports embedding pre-built redirect URL though, so unfortunately I think we'll still need to rely on JavaScript after they press the link.

app/signIn/page.tsx
```diff
@@ -56,15 +63,17 @@ export default function Page() {
       straight to it. Promises explicitly ignored because `handleOAuth` already creates a
       toast for errors
     */
-    if (authConfig.githubEnabled)
-      void handleOAuth('github', setLoggedInUser, authConfig); // skipcq: JS-0098
-    if (authConfig.customEnabled)
-      void handleOAuth('custom', setLoggedInUser, authConfig); // skipcq: JS-0098
+    // redirect();
```
Ryangman: This should be uncommented
```suggestion
    redirect();
```

Ryangman (changes requested):
Looks good! 1 minor bug remains from testing but other than that it's ready to mainline

compose.test.yaml
```diff
@@ -31,6 +31,8 @@ services:
         condition: service_healthy
     environment:
       DATABASE_URL: mysql://root:${DB_PASSWORD}@mariadb/${DB_DATABASE}
+      BETTER_AUTH_URL: ${BETTER_AUTH_URL}
+      DISABLE_PASSWORD_STRENGTH_CHECK: true
```
Ryangman: I may be wrong but for test builds I think this should probably be false. I know at least one e2e test needs password strength check enabled to test the toast provided for a compromised password. 

Especially considering you manually set this env var in the package.json script `start:test`

Ryangman: determined correct from meeting

Ryangman (dismissed):
With the compose file clarification this looks good to merge

sabri414 (changes requested):
Thanks, this is looking close. I found two small issues I’d like fixed before approving:
1. `page.tsx` calls `redirect()` during render when local auth is disabled and exactly one SSO provider is configured. Since this is a client component, re-renders can trigger `handleOAuth()` multiple times. Could this be moved into a guarded `useEffect`, keeping the fallback link as the manual retry?
2. `SignUp` still renders its own `<OAuth />` block, while `page.tsx` now renders the shared OAuth block below the tabs. With local auth + SSO enabled, switching to the Sign Up tab shows duplicate provider buttons. I think the OAuth block should live in one place only.
Once those are addressed, I’m happy to approve.

sabri414 (approved):
Thanks for fixing the duplicate OAuth buttons on the Sign Up tab. I still generally prefer keeping side effects like redirects out of render, but given this page immediately unloads into the SSO flow and the auth config is static for the running app, I’m okay treating that as non-blocking here. I re-checked the updated branch and the targeted sign-in/OAuth tests pass. Approving.

Ryangman (approved):
lgtm

PR #155, opened 2026-05-05
feat: allow items to change sections via dropdown menu

### LINKED ISSUE #76: Allow items to change sections

lilyvesstible:
Items can now change sections. This is available via a dropdown menu in the More menu of each item. The dropdown will display which section the item is currently in, and give options to change to a different section.

## Related Issues

- Closes Issue #76 

## Changes

- [X] Adds dropdown menu in the More menu to change the section of an item
- [X] Adds an API endpoint to change the section of an item

### Tests added

- Validate if API is called when item section changes

## Reviewer Notes

There's an error in my test that doesn't stop it from passing the test. I'm not entirely sure what's causing it. I'll bring it up next meeting and resolve it soon.


components/List/listReducer.ts
```diff
@@ -155,86 +113,94 @@ export default function listReducer( // skipcq: JS-0045, JS-R1005
...
-        sectionItems.filter(item => item === action.id)
-      );
+      section.items.set(action.targetItem.id, action.targetItem);
 
-      newState.items.delete(action.id);
+      //Copied from 'DeleteItem'
+      newState.list.sections
+        .get(action.pastSectionId)
+        ?.items.delete(action.targetItem.id);
       break;
```
radiantBear: Please wrap this in a lexical block to resolve SonalCloud's warning

components/ListItem/__tests__/ListItem.test.tsx
```diff
@@ -174,3 +163,43 @@ it('Displays all members assigned to the item', () => {
...
+    </HeroUIProvider>
+  );
+  const user = userEvent.setup();
+
+  await user.click(getByTestId('more-button'));
+  expect(getByTestId('item-section-select')).toHaveTextContent('section1');
+
+  await user.click(getByTestId('item-section-select'));
+  expect(getByTestId('section1-select-item')).toHaveTextContent('section1');
+  expect(getByTestId('section2-select-item')).toHaveTextContent('section2');
```
radiantBear: Rather than using test IDs, please get these elements by role and/or label where possible. There should be an example for retrieving `<More>` in some other test, perhaps for the comment feature?

components/ListItem/ItemSection.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * This is the component used for changing sections within an item's 'more' menu.
+ *
+ * @param currentSection The section this item is currently associated with
+ * @param totalSections A total list of sections in the larger list
+ * @param updateSection The passed function that changes an item's section
+ */
+export default function ItemSection({
+  currentSection,
+  totalSections,
+  updateSection
```
radiantBear: In keeping with React's naming conventions, please start this event callback with `on`, e.g.
```suggestion
  onSectionChange
```

components/List/List.tsx
```diff
@@ -91,34 +101,34 @@ export default function List({
...
-          )}
           listId={list.id}
-          members={stateToMembers(list.members, builtRoles)}
+          members={list.members}
           section={section}
-          tags={list.tags.values().toArray()}
-          onItemChange={dispatchList}
-          onSectionChange={dispatchList}
+          tagsAvailable={tagsAvailable}
+          totalSections={Array.from(list.sections.values()).map(section => [section.id, section.name])}
```
radiantBear: The `Map` type is well-suited for this pattern of mapping IDs to names and is a bit more clear than a tuple

```suggestion
          totalSections={new Map(list.sections.values().map(section => [section.id, section.name])}
```

components/ListItem/ItemSection.tsx
```diff
@@ -0,0 +1,66 @@
...
+  }));
+
+  return (
+    <div className={'-mt-2 -mb-2'}>
+      <Select
+        className={'w-28 grow-0 shrink-0'}
+        data-testid='item-section-select'
+        items={selectItems}
+        label={<span className='ml-2 text-foreground'>Section</span>}
+        placeholder={currentSection[1]}
```
radiantBear: Why is the current selection being used as a placeholder? It should probably be in `selection` (`selectedItems`? I can't quite remember the name) instead

components/ListItem/ItemSection.tsx
```diff
@@ -0,0 +1,66 @@
...
+        onChange={updateSection}
+      >
+        {selectItems => (
+          <SelectItem
+            key={selectItems.id}
+            data-testid={`${selectItems.label}-select-item`}
```
radiantBear: Instead of setting a `testid` here, it would be better to access this item by using a label name selector on the parent `<Select>`

components/ListItem/More.tsx
```diff
@@ -157,16 +159,25 @@ export default function More({
                       onValueChange={itemHandlers.setDueDate}
                     />
                   ) : null}
+                  <ItemSection
+                    currentSection={
+                      currentSection || ['NO ID AVAILABLE', 'NO SECTION INFO']
+                    }
+                    totalSections={
+                      totalSections || [['NO ID AVAILABLE', 'NO SECTION INFO']]
+                    }
```
radiantBear: Why are `totalSections` and `currentSection` maybe undefined? This shouldn't ever be possible (the database doesn't allow `sectionId` to be `NULL`), so these default cases shouldn't be needed.

components/ListItem/types.ts
```diff
@@ -35,13 +34,15 @@ import { itemHandlerFactory } from './handlerFactory';
...
-  members: Omit<ListMember, 'role'>[];
-  tags: Tag[];
-  hasTimeTracking: FullState['hasTimeTracking'];
-  hasDueDates: FullState['hasDueDates'];
+  list?: List;
+  members: ListMember[];
+  tagsAvailable: Tag[];
+  hasTimeTracking: boolean;
+  hasDueDates: boolean;
+  currentSection?: [string, string];
```
radiantBear: Duplicated data should be minimized as much as possible; only the current section's ID should be stored here because the name can be looked up from that ID. If `totalSections` is also changed to a `Map`, this is an $O(1)$ operation
```suggestion
  sectionId: string;
```

components/ListSection/SectionBody.tsx
```diff
@@ -49,36 +48,43 @@ import { ItemAction } from '../List';
...
+ * @param totalSections A total list of sections in the larger list
+ * @param dispatchItemChange Callback for updating an item's state
+ * @param reorderItem Callback for finalizing the new order of items and updating React
+ *  state
  * @param addNewTag Callback to propagate state changes when a new tag is created from the
  *  "add tag" menu
- * @param onItemEvent Callback for updating an item's state
- * @param onItemReorder Callback for finalizing the new order of items and updating React
- *  state
+ * @param updateSection The passed function that changes an item's section
```
radiantBear: It doesn't look like this is actually passed in
```suggestion
```

components/ListItem/ItemSection.tsx
```diff
@@ -0,0 +1,66 @@
...
+  );
+  const selectItems = updatedTotalSections.map(([id, label]) => ({
+    id,
+    label
+  }));
+
+  return (
+    <div className={'-mt-2 -mb-2'}>
+      <Select
+        className={'w-28 grow-0 shrink-0'}
```
radiantBear: `w-28` is too narrow for the section dropdown; even short section names are getting cut off
```suggestion
        className='w-full grow-0 shrink-0'
```
Also, I think `variant='underlined'` would fit the style better; what do you think?

components/ListItem/More.tsx
```diff
@@ -157,16 +159,25 @@ export default function More({
                       onValueChange={itemHandlers.setDueDate}
                     />
                   ) : null}
+                  <ItemSection
```
radiantBear: Could you please move `<ItemSection>` outside the `<div>` wrapping `<Priority>` and `<DateInput>`? Section names can be kinda long, so it'd be good to give the dropdown a full row to minimize truncation.

radiantBear (changes requested):
Thanks Lily! This is a great start for allowing section change. I was recently looking through my very first roadmap for the project and this is one of the main unfinished items from that, so it's great to finally see it working! Below are a few notes on implementation details, but the broad strokes are definitely there.

components/ListItem/More.tsx
lilyvesstible: For some reason, it's expecting both of these values to possibly be null. I don't know why it's doing this, cause I never say it might be null anywhere. It's also running just fine if I don't have this undefined case, so should I just get rid of it?

components/ListItem/types.ts
lilyvesstible: How do I get the section name from an id?

lib/database/list.ts
```diff
@@ -393,4 +394,45 @@ export async function querySectionInList(
   });
 
   return Boolean(result);
+ * Revokes a list member's access and unassigns them from all items in the list
+ *
+ * @param listId The list to remove a member from
+ * @param userId The user to revoke access for
+ * @returns Whether the database accepted the update
+ */
```
radiantBear, line 438: This is a syntax error. The previous function is not closed and this new comment has no start. This function doesn't seem to have anything to do with your work, so maybe originated from a merge conflict with `main`?

app/api/item/[id]/route.ts
```diff
@@ -65,6 +65,24 @@ export async function PATCH(
...
+      item.elapsedMs = requestBody.elapsedMs;
+    if (requestBody.dateStarted !== undefined)
+      item.dateStarted = requestBody.dateStarted
+        ? new Date(requestBody.dateStarted)
+        : null;
+    if (requestBody.dateCompleted !== undefined)
+      item.dateCompleted = requestBody.dateCompleted
+        ? new Date(requestBody.dateCompleted)
+        : null;
+
```
Comment (radiantBear) [inline app/api/item/[id]/route.ts]: This block of changes is intentionally excluded from the `if` branch. Only the section **OR** other fields can be modified; not both.

components/ListItem/ItemSection.tsx
```diff
@@ -21,22 +21,24 @@ import { Select, SelectItem } from '@heroui/react';
...
-  updateSection
+  sectionId,
+  onUpdateSection
 }: {
-  currentSection: [string, string];
-  totalSections: [string, string][];
-  updateSection: (e: unknown) => unknown;
+  totalSections: Map<string, string>;
+  sectionId: string;
+  onUpdateSection: (e: never) => unknown;
```
radiantBear: The type of `e` isn't `never`; it's some form event passed by HeroUI's `<Select>`'s `onChange` callback. 

As a side note, you might find it easier to use `onSelectionChange` - there's examples of that scattered through the codebase if you decide to switch.

components/ListItem/ItemSection.tsx
```diff
@@ -21,22 +21,24 @@ import { Select, SelectItem } from '@heroui/react';
...
 }) {
-  const updatedTotalSections = totalSections.filter(
-    ([name]) => name !== currentSection[1]
-  );
+  const currentSection = totalSections.get(sectionId);
```
radiantBear:
```suggestion
  const currentSectionName = totalSections.get(sectionId);
```

components/ListItem/ItemSection.tsx
```diff
@@ -45,17 +47,18 @@ export default function ItemSection({
   return (
     <div className={'-mt-2 -mb-2'}>
       <Select
-        className={'w-28 grow-0 shrink-0'}
-        data-testid='item-section-select'
+        aria-label='item-section-select'
```
radiantBear: Aria labels should be natural English (for use by screen readers), for example `'Select a section to move this item to'`.

components/ListItem/ItemSection.tsx
```diff
@@ -45,17 +47,18 @@ export default function ItemSection({
...
-        onChange={updateSection}
+        selectedKeys={[sectionId || 'No Section']}
+        variant='underlined'
+        onChange={onUpdateSection}
       >
         {selectItems => (
           <SelectItem
             key={selectItems.id}
-            data-testid={`${selectItems.label}-select-item`}
+            aria-label={`${selectItems.label}-select-item`}
```
radiantBear: Please make a similar change to this `aria-label` (or remove it entirely since this element's contents are already a natural label for it)

components/ListSection/SectionBody.tsx
```diff
@@ -48,43 +49,41 @@ import { ItemAction } from '../List';
...
- * @param currentSection The section this item is currently associated with
  * @param totalSections A total list of sections in the larger list
- * @param dispatchItemChange Callback for updating an item's state
- * @param reorderItem Callback for finalizing the new order of items and updating React
- *  state
  * @param addNewTag Callback to propagate state changes when a new tag is created from the
  *  "add tag" menu
- * @param updateSection The passed function that changes an item's section
+ * @param addNewTag Callback to propagate state changes when a new tag is created from the
+ *  "add tag" menu
```
radiantBear: Looks like this is duplicated now
```suggestion
```

lib/database/list.ts
```diff
@@ -394,7 +394,9 @@ export async function querySectionInList(
   });
 
   return Boolean(result);
- * Revokes a list member's access and unassigns them from all items in the list
+}
+
+/* Revokes a list member's access and unassigns them from all items in the list
  *
```
radiantBear:
```suggestion
/**
 * Revokes a list member's access and unassigns them from all items in the list
 * 
```

components/List/listReducer.ts
```diff
@@ -113,94 +159,117 @@ export default function listReducer(
...
+      if (!targetSectionItems)
         throw new Error(
           `Unable to find section with ID ${action.targetSectionId}`
         );
 
-      section.items.set(action.targetItem.id, action.targetItem);
+      targetSectionItems.push(action.targetItem.id);
+      newState.items.set(action.targetItem.id, action.targetItem);
+      newState.itemAssignees.set(action.targetItem.id, []);
+      newState.itemTags.set(action.targetItem.id, []);
```
radiantBear: Why are the item's tags and assignees cleared when it's moved between sections?

components/List/listReducer.ts
```diff
@@ -113,94 +159,117 @@ export default function listReducer(
...
 
-      if (!section)
+      if (!targetSectionItems)
         throw new Error(
           `Unable to find section with ID ${action.targetSectionId}`
         );
 
-      section.items.set(action.targetItem.id, action.targetItem);
+      targetSectionItems.push(action.targetItem.id);
+      newState.items.set(action.targetItem.id, action.targetItem);
```
radiantBear: I'm not a fan of updating the whole item from the `ChangeItemSection` action. This opens the door for unexpected changes to other item state from this action. It appears that this isn't even needed since `ListItem.tsx:215` just passes the current item in, so this should probably be removed altogether.

If the whole item isn't needed, it would also be better to change this action to only take a `targetItemId` instead of the whole item.

components/List/listReducer.ts
```diff
@@ -113,94 +159,117 @@ export default function listReducer(
...
+
+      if (!pastSectionItems)
+        throw new Error(
+          `Unable to find items for section ${action.pastSectionId}`
+        );
+
+      newState.sectionItems.set(
+        action.pastSectionId,
+        pastSectionItems.filter(item => item !== action.targetItem.id)
+      );
```
radiantBear: `pastSectionItems` also need to be updated so there isn't a gap in their `sectionIndex`s (i.e. any `sectionIndex` greater than `targetItem.sectionIndex` need to be decremented)

radiantBear (commented):
Thank you Lily! Here are some follow-on thoughts:

app/api/item/[id]/route.ts
Comment (lilyvesstible) [reply app/api/item/[id]/route.ts]: Are you saying to move this to the else branch? So that sectionID has to be unmodified?

components/ListItem/ItemSection.tsx
```diff
@@ -58,7 +58,7 @@ export default function ItemSection({
         {selectItems => (
           <SelectItem
             key={selectItems.id}
-            aria-label={`${selectItems.label}-select-item`}
+            aria-label={`Select ${selectItems.label}`}
```
radiantBear:
```suggestion
            aria-label={`Move to ${selectItems.label}`}
```

app/api/item/[id]/route.test.ts
```diff
@@ -52,7 +55,10 @@ beforeEach(() => {
 describe('PATCH', () => {
   test('Updates item name when provided & requestor has permissions', async () => {
     vi.mocked(getUser).mockResolvedValue(MOCK_USER);
-    vi.mocked(getListItemById).mockResolvedValue(MOCK_ITEM);
+    vi.mocked(getListItemById).mockResolvedValue({
+      ...MOCK_ITEM,
+      listId: 'dummy-id'
+    });
```
Comment (radiantBear) [inline app/api/item/[id]/route.test.ts]: Every test that sets `MOCK_ITEM` as the resolved value adds the `listId` field, too. Instead of putting that everywhere, the `MOCK_ITEM` definition should be updated to add a `listId` field:

```ts
const MOCK_ITEM = { ...new ListItem('test tag', 'dummySection', {}), listId: 'dummy-id' };
```

app/api/item/[id]/route.test.ts
```diff
@@ -167,79 +215,52 @@ describe('PATCH', () => {
...
-    test('Rejects unauthenticated users', async () => {
-      vi.mocked(getUser).mockResolvedValue(false);
-
-      const response = await DELETE(
-        new Request(ITEM_PATH, {
-          method: 'delete'
-        }),
-        {
-          params: Promise.resolve({ id: MOCK_ITEM.id })
-        }
```
Comment (radiantBear) [inline app/api/item/[id]/route.test.ts:206]: Why were all the `DELETE` route's tests removed?

Comment (lilyvesstible) [reply app/api/item/[id]/route.test.ts:206]: I think the DELETE tests were moved to the bottom. I think these are the same, but let me know if anything needs to be fixed from this
Screenshot 2026-05-20 154113

Comment (radiantBear) [reply app/api/item/[id]/route.test.ts:206]: Wow, that's unfortunate! I thought `git`'s diffs were more powerful than that.

app/api/item/[id]/route.test.ts
Comment (radiantBear) [inline app/api/item/[id]/route.test.ts]: Now that `MOCK_ITEM` specifies a `listId`, these don't need to add it also

```suggestion
    vi.mocked(getListItemById).mockResolvedValue(MOCK_ITEM);
```

radiantBear (dismissed):

radiantBear (dismissed):

lib/transformations/list/stateToState.ts
```diff
@@ -248,6 +249,49 @@ function sectionReducer<
...
+      );
+
+      if (!targetSectionItems)
+        throw new Error(
+          `Unable to find section with ID ${action.targetSectionId}`
+        );
+
+      const newItem = getItem(newState, action.targetItemId);
+
+      newItem.sectionId = action.targetSectionId;
```
sabri414, line 387: When changing sections, this updates the item’s sectionId but not its sectionIndex. The backend appends the item to the target section, so the frontend state should also set the moved item’s index to the end of the target section. Otherwise, the UI order can disagree with the database until refresh.

components/ListItem/ItemSection.tsx
```diff
@@ -0,0 +1,69 @@
...
+}: {
+  totalSections: Map<string, string>;
+  sectionId: string;
+  onUpdateSection: (e: React.ChangeEvent<HTMLSelectElement>) => unknown;
+}) {
+  const currentSectionName = totalSections.get(sectionId);
+
+  const updatedTotalSections = totalSections
+    .entries()
+    .filter(([name]) => name !== currentSectionName);
```
sabri414: This filter compares the section id/key to the current section name, so the current section is not actually filtered out. The dropdown still offers “move to the section I’m already in.” This should filter by id, probably ([id]) => id !== sectionId.

lib/transformations/list/stateToState.ts
```diff
@@ -248,6 +249,49 @@ function sectionReducer<
...
+      newState.items.set(action.targetItemId, newItem);
+
+      targetSectionItems.push(action.targetItemId);
```
sabri414, line 391: Related edge case: if the current section remains selectable and the user selects it, targetSectionItems and pastSectionItems can be the same section. This push/filter flow can remove the item from the section in local state. Filtering out the current section should prevent that path.

app/api/item/[id]/route.ts
```diff
@@ -58,27 +59,40 @@ export async function PATCH(
...
-  if (requestBody.dateCompleted !== undefined)
-    item.dateCompleted = requestBody.dateCompleted
-      ? new Date(requestBody.dateCompleted)
-      : null;
-
-  const result = await updateListItem(item);
-
-  if (!result) return ServerError.Internal('Could not update item');
+  if (requestBody.sectionId) {
+    const allowed = await querySectionInList(
```
Comment (sabri414) [inline app/api/item/[id]/route.ts:80]: Because this branches on sectionId, a PATCH body with sectionId plus another field, like name or description, returns success but ignores the other fields. Could we either reject mixed updates with a 400 or apply the normal item field updates as well?

sabri414 (changes requested):
Thanks for fixing the merge conflict.

components/ListItem/ItemSection.tsx
radiantBear: Oh, good catch! Instead of comparing names at all, it would be better to compare IDs since multiple sections *could* have the same name

app/api/item/[id]/route.ts
Comment (radiantBear) [reply app/api/item/[id]/route.ts:80]: @Ryangman rejecting mixed updates seems reasonable...

Comment (Ryangman) [reply app/api/item/[id]/route.ts:80]: Rejecting bodies with a sectionId and anything else via zod rules seems like a fine solution to me. Applying normal field item updates as well is not a reasonable solution, since there will inherently be two database requests that would have to be serialized to provide any confidence that both succeed/fail.

components/ListItem/ItemSection.tsx
lilyvesstible: After thinking about it for a bit, I think it would be best if there were no filter, instead purposefully displaying the current section. I have some reasons:
1. HeroUI indicates which option is currently selected, so there wouldn't be any confusion on which section the item is currently in, or multiple sections with the same name
2. If I remove the current section from the selection, I can't show within the selection which section is currently selected (what a tongue twister lol). 

Let me know what you guys think of this

lib/transformations/list/stateToState.ts
lilyvesstible: Related to above, because I opted to keep the current section in the selection, I also added a check within the changeItemSection function to see if the current section is the same as the target section. Hopefully this is enough to stop this edge case from happening on an api level.

app/api/item/[id]/route.ts
Comment (lilyvesstible) [reply app/api/item/[id]/route.ts:80]: What should I do to make this change? Or alternatively, could one of you enact this change? I apologize, I'm not very familiar with how this works, and you guys seem to have a good idea of what needs to happen.

components/ListItem/ItemSection.tsx
radiantBear: That sounds good for me, thanks for thinking about it more Lily!

sabri414 (commented):
Thanks for the update. I think this is closer, but I’m still not ready to approve yet.
The section move state still looks slightly off to me. The moved item now gets its new index from the target section, which is good. But after that, the reducer uses that new target index when deciding which items in the old section should have their indexes decremented. I think it needs to keep the item’s original index from the old section first, then use that original index to close the gap in the old section.
Example: if an item moves from index 1 in its old section into a target section that already has 5 items, the moved item’s new index becomes 5. Then the old section cleanup checks for items with an index greater than 5, so items after old index 1 do not get shifted down. That can leave a gap in the frontend state.
Also, I think Ryan’s suggestion about mixed PATCH bodies still needs to be handled. If a request includes sectionId plus another field like name, the route still returns success but ignores the other field. It sounds like the preferred fix is to reject those mixed bodies with Zod.

radiantBear (dismissed):

lib/model/listItem.ts
```diff
@@ -64,9 +64,11 @@ export default class ListItem {
   tags: Tag[];
   issueId: bigint | null;
   listId?: string;
+  sectionId: string;
 
   constructor(
     name: string,
+    sectionId: string,
     {
```
sabri414, line 76: Since this changes the `ListItem` constructor to require `sectionId` as the second argument, I think there is one call site from the newly merged GitHub webhook code that still needs updating. `github/webhook.ts` currently calls: `new ListItem(name, { priority, description, dateCreated, issueId })`. That file may not show up in this PR’s “Files changed” tab because it came from `main`, but after this constructor change it will pass the options object where the section id should go.

Could we update that webhook path so the section exists first, then call:

`new ListItem(name, list.sectionId, { priority, description, dateCreated, issueId })`

lilyvesstible: Yeah i think this was an issue that existed in main, cause my other pr is giving that error. Still, I managed to make some fixes to it (i think)

lib/transformations/list/stateToState.ts
```diff
@@ -368,6 +370,50 @@ function sectionReducer<
...
+      if (!targetSectionItems)
+        throw new Error(
+          `Unable to find section with ID ${action.targetSectionId}`
+        );
+
+      const newItem = getItem(newState, action.targetItemId);
+      const oldIndex = newItem.sectionIndex;
+
+      newItem.sectionIndex = targetSectionItems.length;
+      newState.items.set(action.targetItemId, newItem);
```
sabri414, line 389: The reducer updates the moved item’s `sectionIndex`, but it no longer updates `newItem.sectionId` to `action.targetSectionId`. That can leave client state saying the item still belongs to its old section after the move.

lib/database/listItem.ts
```diff
@@ -220,3 +237,51 @@ export async function unlinkAssignee(
...
+        await tx.item.update({
+          where: { id: item.id },
+          data: {
+            sectionId,
+            sectionIndex: targetSectionItems
+          }
+        });
+        await tx.item.updateMany({
+          where: {
+            sectionId,
```
sabri414: updateItemSection reindexes using the new target sectionId. Since the item has just been moved, this decrements items in the destination section instead of closing the gap in the old/source section. I think this should use the item’s original sectionId.

sabri414 (changes requested):
Thanks for the updates here. The webhook constructor issue looks fixed now, and the PR is mergeable with green checks. I think there are still two section-move bugs to fix before I can approve. Once those two are fixed, I think this is very close.

sabri414 (dismissed):
Thanks for sticking with this and addressing the review feedback. The previous issues I was concerned about look fixed now

lib/database/listItem.ts
```diff
@@ -251,7 +266,7 @@ export async function updateItemSection(
         });
         await tx.item.updateMany({
           where: {
-            sectionId: item.sectionId,
+            sectionId: originalSectionId,
```
radiantBear: Well, here's a "too many cooks in the kitchen" problem :).

Sabri is incorrect that `item.sectionId` refers to the item's new section ID on line 269. Lines 260-266 update the `sectionId` in the database, but not in JavaScript. This change is unnecessary; please revert it.

sabri414: Yep, that makes sense. I think my concern was about accidentally using the target `sectionId`, but you’re right that `item.sectionId` itself is still the original/source section in JavaScript. The Prisma update changes the database row, not the local `item` object. So I’m good with reverting `originalSectionId` and using `item.sectionId` directly here, as long as we don’t go back to using the target `sectionId`.

lib/database/listItem.ts
```diff
@@ -220,3 +237,51 @@ export async function unlinkAssignee(
...
+            sectionId,
+            sectionIndex: targetSectionItems
+          }
+        });
+        await tx.item.updateMany({
+          where: {
+            sectionId,
+            sectionIndex: {
+              gte: originalIndex
+            }
```
sabri414, line 271: I agree with Nate’s point that we don’t need `originalSectionId`, since `item.sectionId` is still the original/source section in JavaScript.
But I think the latest revert went one step too far: this now uses the bare `sectionId` parameter in the `updateMany` filter, and that parameter is the target section being moved into.

Could we change this to use `item.sectionId` directly?

```ts
where: {
  sectionId: item.sectionId,
  sectionIndex: {
    gte: originalIndex
  }
}

sabri414 (commented):
The other fixes still look good: webhook constructor is fixed, client reducer updates `sectionId`, PR is mergeable, and the latest PR metadata showed green checks.

lib/database/listItem.ts
lilyvesstible: The current code is basically identical to this. sectionId is shorthand in js for sectionId: item.sectionId

radiantBear: `sectionId` is not a shortcut for `sectionId: item.sectionId`; it's a shortcut for `sectionId: sectionId`.

The variable `item.sectionId` holds the old section ID, but the variable `sectionId` holds the new section ID. The old one is needed for the query to shift items up in the old section, so Sabri's change is correct.

radiantBear (dismissed):

lib/database/listItem.ts
lilyvesstible: Ok nevermind I don't know what I was thinking. Made the change now.

sabri414 (dismissed):
Thanks for sticking to this and applying our feedback! Great job Lily

radiantBear (approved):

sabri414 (approved):

Issue #114, opened 2026-03-12
Fix today page


radiantBear:
Data doesn't correctly populate or update for the page

PR #157, opened 2026-05-08
refactor: relocate list state transitions for reuse between components

radiantBear:
Moves functions for generating & updating list state to a more general location so both `<List>` (used on `/list/{id}`) and `<ListGroup>` (used on `/list`) can use them to manage state.

## Related Issues

- Preparation for #114

## Changes

- [x] Relocated functions as stated above
- [x] Changed list actions to require a list ID for use with multiple lists

### Tests fixed
- Incorporated new list action parameter (list ID)

## Reviewer Notes

What do you think about this file organization? I think these functions should be somewhere in `/lib` and allow for future expansion, but this may be a little too complicated of organization.


lilyvesstible (dismissed):

lilyvesstible (dismissed):

A-man37 (approved):
Looks good!

lilyvesstible (approved):

PR #159, opened 2026-05-22
fix: users can't remove tag if not linked to any items

radiantBear:
Fixes a database query bug that makes tags seem like they're not linked to their list if they aren't assigned to at least 1 item.

## Testing

Since this affects the success of database queries, it cannot be tested via unit/integration tests. To validate that users can now delete tags that aren't linked to any items:
- Create a new tag on a list
- Try to delete it from the Settings modal when it's not linked to any items


sabri414 (approved):
Looks good to me

Ryangman (approved):
lgtm

PR #162, opened 2026-05-23
feat: show top 10 items from all lists on "today" page

### LINKED ISSUE #114: Fix today page

radiantBear:
Re-implements "Today" page to display the top list items from across all lists a user is a member of.

## Related Issues

- Closes #114 

## Changes

- [x] Added new reducer function to handle multi-list state
- [x] Added "today" page that grabs all lists the user is a member of
- [x] Added helper for merging `Map` objects (needed for combining lists' state)
- [x] Fixed a few aria labels
- [x] Validated that `<ListItemGroup>` functions as expected

### Tests added

- Tests for utility function merging `Map`s
- Tests for `<ListItemGroup>`'s current behavior


lilyvesstible (dismissed):

components/ListItemGroup/ListItemGroup.tsx
```diff
@@ -0,0 +1,192 @@
...
+        item
+      )}
+      list={itemGroup.lists.get(item.listId)}
+      members={itemGroupStateToMembers(
+        itemGroup.listMembers,
+        itemGroup.members,
+        roles,
+        item.listId
+      )}
+      sectionId='unknown'
```
sabri414, line 183: This needs the real section id rather than `"unknown"`. `ListItem` reuses the existing delete flow, and `deleteSelf()` dispatches `DeleteItem` with this `sectionId` after the API delete succeeds. In the Today view that sends `sectionId: "unknown"` into `itemGroupReducer`, which then looks up `sectionItems.get("unknown")` and throws. Result: deleting an item from Today can successfully delete it server-side and then break/stale the client state. Can we preserve each item's section id in the group state and add a regression test for deleting from Today?

components/ListItemGroup/__tests__/ListItemGroup.test.tsx
```diff
@@ -0,0 +1,341 @@
...
+    });
+    expect(api.post).toHaveBeenCalledWith(
+      '/item/item-id-3/tag/tag-id',
+      expect.any(Object)
+    );
+
+    await user.click(within(item2).getByLabelText('Update tags'));
+    await waitFor(() => expect(getByText('New tag')).toBeVisible());
+    expect(queryByText(MOCK_TAG.name)).not.toBeInTheDocument();
+    screen.debug(undefined, Number.MAX_SAFE_INTEGER, { highlight: false });
```
sabri414: Can we remove this debug dump before merging? It prints a very large DOM tree during the test run, which makes CI/test logs noisy and harder to scan.

sabri414 (commented):
I’d hold approval until CI is green. The PR currently has failing Cypress and DeepSource coverage checks on the latest head I reviewed, and there’s at least one Today-view delete-state bug

components/ListItemGroup/ListItemGroup.tsx
radiantBear: Yes, leaving `sectionId='unknown'` is not a long-term solution. However, efficiently identifying a list item's section requires Lily's PR to be merged first. Since we're down to the wire, I don't think it's worth blocking this feature entirely because of the edge case, but we should definitely create an issue to track this bug once it's merged.

lilyvesstible (dismissed):

sabri414 (approved):
Approving to unblock, since CI is green now and the debug-output issue was cleaned up.
I’m still concerned about the known `sectionId='unknown'` behavior in the Today view. As noted in the thread, deleting an item from Today can succeed server-side and then fail/stale the client state because the reducer does not have the real section id. I understand the reasoning that the clean fix depends on Lily’s PR and that this may be an acceptable edge case for this release.
Please open and link a follow-up issue for that bug so it does not get lost after merge. With that explicitly tracked, I’m okay approving this PR.

lilyvesstible (approved):

PR #163, opened 2026-05-26
fix: prevent unauthenticated access to `/profile` route

### LINKED ISSUE #131: Profile Page allows unauthenticated access
An unauthenticated user can access `/profile` and is not redirected to sign-in via  the `authorize()` function

Ryangman:
Adds authorization check to `/profile` route to redirect unauthenticated users to sign-in, fixing current bug of displaying an empty page.

## Related Issues

Closes #131 

## Changes

- [x] added `authorize()` check at root of `/profile` route
- [x] added tests

## Testing Coverage

Tests added for redirection logic.


app/profile/page.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+  vi.mocked(useAuth).mockReturnValue({
+    authConfig: {
+      localEnabled: true,
+      githubEnabled: false,
+      customEnabled: false
+    },
+    loggedInUser: MOCK_USER,
+    setLoggedInUser: vi.fn()
+  });
+});
```
radiantBear, line 72: It doesn't look like `useAuth` is actually used in the page's source, so it shouldn't need to be mocked here

app/profile/page.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+describe('Redirects for [un]authenticated users', () => {
+  it('renders profile page for authenticated users', async () => {
+    vi.mocked(getUser).mockResolvedValue(MOCK_USER);
+    const resolvedPage = await Page();
+    const { getByRole } = render(
+      <HeroUIProvider disableRipple>{resolvedPage}</HeroUIProvider>
+    );
```
radiantBear, line 80: Is there a reason to not use TSX for all rendered content?
```suggestion
    const { getByRole } = render(
      <HeroUIProvider disableRipple>
        <Page />
      </HeroUIProvider>
    );
```

radiantBear (changes requested):
Thanks Ryan! Just a couple minor improvements I think could be made:

app/profile/page.test.tsx
Ryangman: It is used by the `<AuthSettings>` component included in the page

app/profile/page.test.tsx
Ryangman: That was the method I originally tried, but it results in an almost empty DOM, I believe to be due to this page now being an async server component. 

Image

radiantBear (approved):

app/profile/page.test.tsx
radiantBear: Interesting; it would probably be helpful to document this for future reference

lilyvesstible (approved):

PR #164, opened 2026-05-28
feat(ui): allow lists to be viewed in a kanban layout

### LINKED ISSUE #10: Add board viewing style for tasks

lilyvesstible:
Adds the option for users to view their list in Kanban mode, which displays sections in a horizontal row.

## Related Issues

Closes Issue #10, Partially blocked by Issue #103 

## Changes

- [X] Added switch in General Settings to change Kanban setting.
- [X] Added visual change to lists/sections when Kanban mode is active

### Tests added

- Test for the Kanban setting switch
- Tests for tracking kanban status within lists
- Test for tracking kanban status within sections

## Reviewer Notes

Currently, the sections don't appear as intended because they're so thin. Primary features such as adding a task or checking a section's settings are blocked when in kanban mode. This should be fixed when PR #99 gets merged with main.


components/List/__tests__/List.test.tsx
```diff
@@ -223,6 +237,42 @@ describe('List state changes', () => {
 
     expect(getByLabelText('Auto-order list items')).not.toBeChecked();
   });
+
+  //PUT KANBAN TEST RIGHT HERE!!!!!
```
radiantBear:
```suggestion
```

components/List/__tests__/List.test.tsx
```diff
@@ -223,6 +237,42 @@ describe('List state changes', () => {
...
+        <List
+          startingList={JSON.stringify({ ...MOCK_LIST, isAutoOrdered: true })}
+          startingRoles='[]'
+        />
+      </HeroUIProvider>
+    );
+
+    expect(localStorageMock.getItem).toHaveBeenCalledOnce();
+    expect(localStorageMock.setItem).toHaveBeenCalledOnce();
+    expect(localStorageMock.setItem).toHaveBeenCalledWith(
```
radiantBear: These can be merged with `toHaveBeenCalledExactlyOnceWith`

components/List/__tests__/List.test.tsx
```diff
@@ -223,6 +237,42 @@ describe('List state changes', () => {
...
+        <List
+          startingList={JSON.stringify({ ...MOCK_LIST, isAutoOrdered: true })}
+          startingRoles='[]'
+        />
+      </HeroUIProvider>
+    );
+
+    expect(localStorageMock.getItem).toHaveBeenCalledOnce();
+    expect(localStorageMock.getItem).toReturnWith('[["list-id",true]]');
+    expect(getByTestId('kanban-list-format')).toBeVisible();
```
radiantBear, line 268: Adding an assertion like this to the other test, too, would be good

components/List/__tests__/List.test.tsx
```diff
@@ -223,6 +237,42 @@ describe('List state changes', () => {
 
     expect(getByLabelText('Auto-order list items')).not.toBeChecked();
   });
+
+  //PUT KANBAN TEST RIGHT HERE!!!!!
+  test('List initiates as Kanban being false', () => {
+    render(
+      <HeroUIProvider disableRipple>
+        <List
+          startingList={JSON.stringify({ ...MOCK_LIST, isAutoOrdered: true })}
```
radiantBear: Does auto-ordering matter for this?

```suggestion
          startingList={JSON.stringify(MOCK_LIST)}
```

components/ListSettings/ListSettings.tsx
```diff
@@ -56,12 +56,16 @@ import TagSettings from './TagSettings';
 export default function ListSettings({
   list,
   roles,
+  isKanban,
+  setIsKanban,
```
radiantBear: In keeping with our naming conventions, this callback should start with `on`, e.g. `onKanbanToggle`

components/ListSettings/GeneralSettings.tsx
```diff
@@ -137,6 +163,9 @@ export default function GeneralSettings({
         >
           Auto-order list items
         </Switch>
+        <Switch isSelected={isKanban} size='sm' onValueChange={updateKanban}>
+          View in Kanban Mode
```
radiantBear:
```suggestion
          View in kanban mode
```

components/ListSettings/GeneralSettings.tsx
```diff
@@ -49,15 +49,19 @@ export default function GeneralSettings({
   hasDueDates,
   hasTimeTracking,
   isAutoOrdered,
-  setListName
+  isKanban,
+  setListName,
+  setIsKanban
```
radiantBear: In keeping with our style guide, these callback functions should start with `on`, e.g. `onListNameChange`; `onKanbanToggle`

components/ListSettings/GeneralSettings.tsx
```diff
@@ -104,9 +108,31 @@ export default function GeneralSettings({
...
+   * Gets the current kanban preference for this list, then updates it with a new value.
+   *
+   * @param value The value the preference should update to
+   */
+  function updateKanban(value: boolean) {
+    if (value === isKanban) return;
+
+    const kanbanPref = localStorage.getItem('KanbanPreference') ?? '[]';
+    const kanbanPrefParsed = JSON.parse(kanbanPref) as [string, boolean][];
+    const kanbanPrefMap = new Map<string, boolean>(kanbanPrefParsed);
```
radiantBear: This opens up a possibility for discrepancy between component state and `localStorage` state. I think it would be better to guard this function with `if (value === kanbanPrefMap.get(listId)) return;`

components/List/List.tsx
```diff
@@ -87,50 +87,128 @@ export default function List({
...
+    kanbanPrefMap.set(list.id, false);
+    localStorage.setItem(
+      'KanbanPreference',
+      JSON.stringify(Array.from(kanbanPrefMap.entries()))
+    );
+  }
+
+  const [isKanban, setIsKanban] = useState<boolean>(
+    kanbanPrefMap.get(list.id) ?? false
+  );
```
radiantBear: Ideally, interaction with Kanban preference (for state & localStorage) should be extracted to a React hook to keep `localStorage` logic co-located

components/ListSettings/__tests__/GeneralSettings.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  vi.mocked(useRouter).mockReturnValue(routerMock);
+
+  expect(getByLabelText('View in Kanban Mode')).toBeVisible();
+  await user.click(getByLabelText('View in Kanban Mode'));
+  expect(localStorageMock.setItem).toHaveBeenCalledOnce();
+  expect(localStorageMock.setItem).toHaveBeenCalledWith(
+    'KanbanPreference',
+    '[["listid",true]]'
+  );
```
radiantBear: It would be good to also assert that `setIsKanban` gets called to ensure Kanban mode is activated without refreshing the page

components/List/List.tsx
```diff
@@ -87,50 +87,128 @@ export default function List({
...
+
+  if (isKanban) {
```
radiantBear: Like SonarQube points out, this is a lot of duplication. It looks like the only difference is the second `<span>` in the `true` branch; could you just make this always present but only have styles when `isKanban`?

components/List/List.tsx
```diff
@@ -87,50 +87,128 @@ export default function List({
...
+            onListNameChange={listHandlers.setName}
+          />
+        </span>
+
+        <span
+          className='flex gap-4 items-start overflow-x-scroll'
+          data-testid={'kanban-list-format'}
+        >
+          {Array.from(list.sections.values()).map(section => (
+            //<div key={section.id} className='w-100'>
```
radiantBear:
```suggestion
```

components/ListSection/__tests__/ListSection.test.tsx
```diff
@@ -279,3 +286,28 @@ describe('Section expansion/collapse', () => {
...
+        name: 'Section Name'
+      }}
+      tags={[]}
+      onItemChange={vi.fn()}
+      onSectionChange={vi.fn()}
+      onTagCreate={() => Promise.resolve('')}
+    />
+  );
+
+  expect(getByTestId('kanban-section-format')).toBeVisible();
```
radiantBear, line 321: Instead of just asserting that this `div` exists, it would be good to assert that there's a wrapper element with styles that support Kanban display

components/ListSection/ListSection.tsx
```diff
@@ -108,79 +110,158 @@ export default function ListSection({
...
-              onSectionChange({ type: 'DeleteSection', id: section.id })
-            }
-          />
-        </span>
+  if (isKanban) {
+    return (
+      <div
+        className='rounded-md overflow-hidden border-2 border-content3 box-border shrink-0 shadow-lg shadow-content2 w-100'
+        data-testid={'kanban-section-format'}
+      >
```
radiantBear: This is also a lot of duplication when these props are the only thing that changes between branches. Please merge them together and use e.g. a template string to handle the different styles:

```tsx
<div
  className={`rounded-md overflow-hidden border-2 border-content3 box-border shrink-0 shadow-lg shadow-content2 ${isKanban ? 'w-100' : 'w-full'}`}
  data-testid='section-wrapper'
>
```

radiantBear (changes requested):
Thanks Lily! I'm excited to see this come together!

components/List/__tests__/List.test.tsx
lilyvesstible: Is there a "toHaveReturnedExactlyOnceWith" assertion? I can't find any in the documentation. Also, I do think checking if something was called once, and checking what it returned with are different enough to warrant them being separate. What do you think?

components/List/List.tsx
lilyvesstible: Where should I store this custom hook?

components/List/__tests__/List.test.tsx
radiantBear: Yeah, I think this is a fine assertion! Though not strictly necessary since the mocked function's return value is forced earlier in the test. I don't think there's a way to condense it.

Sorry my initial comment wasn't clear enough - I meant that adding a `expect(getByTestId(...)).not.toBeInTheDocument()` on line 254 would be good.

components/List/List.tsx
radiantBear: You could just put it in a new file in `components/List/` for now!

components/List/__tests__/List.test.tsx
```diff
@@ -263,15 +261,18 @@ describe('List state changes', () => {
     const { getByTestId } = render(
       <HeroUIProvider disableRipple>
         <List
-          startingList={JSON.stringify({ ...MOCK_LIST, isAutoOrdered: true })}
+          startingList={JSON.stringify({ ...MOCK_LIST })}
```
radiantBear: If `MOCK_LIST` is the only thing in this object, there's no need to spread it:
```suggestion
          startingList={JSON.stringify(MOCK_LIST)}
```

components/ListSettings/GeneralSettings.tsx
radiantBear: Please don't forget to update `setListName`!

components/List/List.tsx
```diff
@@ -87,50 +87,84 @@ export default function List({
...
         />
-      ))}
-
-      <AddListSection
-        listId={list.id}
-        onSectionAdded={section =>
-          dispatchList({ type: 'AddSection', id: list.id, section })
-        }
-      />
+      )}
```
radiantBear: It looks like this code is identical to the `<AddListSection` component above, except for the position inside vs outside the `<span>`. Since the `<span>` has no styling rules when `!isKanban`, would it be equivalent to just always use the one inside the span?

lilyvesstible: There is a little visual difference, in that the add section button gets treated like a list item, so the spacing is off. This is what it looks like. Do you think I should just add padding to the AddListSection component or something, or is it ok to have them appear in two separate places like this?
image

radiantBear: The same problem occurs when there's more than 1 section in the list:
image
The `flex gap-4` classes should be unconditionally added to the `<span>` to ensure the spacing remains present in vertical mode. Once that's done, you should be able to deduplicate `<AddListSection>`

components/List/useKanbanPref.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if (!kanbanPrefMap.has(listId)) {
+    kanbanPrefMap.set(listId, false);
+    localStorage.setItem(
+      'KanbanPreference',
+      JSON.stringify(Array.from(kanbanPrefMap.entries()))
+    );
+  }
+
+  return useState<boolean>(kanbanPrefMap.get(listId) ?? false);
```
radiantBear: Instead of returning the `useState`'s update function directly, I'd love to see this hook expanded to automatically handle `localStorage` syncing when the callback is run. That way, consumers of the component won't have to worry about keeping them in sync and all logic related to `localStorage` will be co-located for easier maintenance

components/List/List.tsx
```diff
@@ -120,7 +106,7 @@ export default function List({
       </span>
 
       <span
-        className={isKanban ? 'flex gap-4 items-start overflow-x-scroll' : ''}
+        className={`flex gap-4 ${isKanban ? 'items-start overflow-x-scroll' : 'flex-col'}`}
```
radiantBear: My bad, this should be `gap-8` in non-Kanan mode. Also, could you please change `overflow-x-scroll` to `overflow-x-auto` so this scrollbar space doesn't show up if scrolling isn't needed?

Image

components/List/List.tsx
```diff
@@ -87,50 +88,59 @@ export default function List({
...
-          members={listStateToMembers(list.members, builtRoles)}
-          section={section}
-          tags={list.tags.values().toArray()}
-          onItemChange={dispatchList}
-          onSectionChange={dispatchList}
-          onTagCreate={listHandlers.addNewTag}
+          onSectionAdded={section =>
+            dispatchList({ type: 'AddSection', listId: list.id, section })
+          }
         />
```
radiantBear: Thanks for merging these `<AddListSection>`s together! Could you please also wrap it in a `<span>` so it stays the width of 1 column in Kanban mode instead of stretching like this?

Image

Or do you think it looks better stretched?

components/List/useKanbanPref.ts
lilyvesstible: I think I'm a little confused. From my understanding, the code already handles localStorage syncing whenever its run (getting the local storage, and then setting it to false if its the first time being accessed).

Do you mean having a separate hook for changing kanban status? like a useViewKanbanPref and a useSetKanbanPref?

radiantBear: I mean instead of returning `setIsKanban` from `useState`, it'd be good for the hook to return a function that updates `localStorage` and not just React state. I think that logic is in `GeneralSettings.tsx` right now?

components/List/useKanbanPref.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Returns the current kanban preference for the list, and also returns a function to update the kanban preference
+ *
+ * @param listId The list ID of the kanban preference
+ */
+
+import { useState } from 'react';
+
+export function useKanbanPref(listId: string) {
+  const kanbanPref = localStorage.getItem('KanbanPreference') ?? '[]';
```
sabri414: I’d like this fixed before approving: `useKanbanPref` reads/writes `localStorage` during render. That makes rendering impure and can break outside the browser or mutate storage during aborted/retried renders. Please move the storage read behind a browser guard/lazy effect and only write from the toggle handler or an effect.

sabri414 (changes requested):

components/List/useKanbanPref.ts
```diff
@@ -0,0 +1,89 @@
...
+  };
+};
+
+/**
+ * Returns the current kanban preference for the list, and also returns a function to update the kanban preference
+ *
+ * @param listId The list ID of the kanban preference
+ */
+export function useKanbanPref(listId: string) {
+  const kanbanPref = useSyncExternalStore(subscribe, getSnapshot);
```
sabri414: `useSyncExternalStore` needs a third `getServerSnapshot` argument here. Client components can still be rendered on the server, and this currently throws during server render with `Missing getServerSnapshot`. Something like `useSyncExternalStore(subscribe, getSnapshot, () => '[]')` would make the hook safe to render before `localStorage` exists.

components/List/useKanbanPref.ts
```diff
@@ -0,0 +1,89 @@
...
+  const kanbanPrefParsed = JSON.parse(kanbanPref) as [string, boolean][];
+  const kanbanPrefMap = new Map<string, boolean>(kanbanPrefParsed);
+
+  if (!kanbanPrefMap.has(listId)) {
```
sabri414: This still mutates `localStorage` during render when the list has no saved preference. Render should stay pure because React can retry, abort, or double-render components. Could we avoid storing default `false` entries here and just return `false` when the map is missing this `listId`? Then only write to `localStorage` from `setIsKanban` when the user actually toggles the setting.

sabri414 (commented):
Thanks for the update. The hook is closer, but I still think this needs changes before approval: `useSyncExternalStore` is missing `getServerSnapshot`, and the hook still writes to `localStorage` during render for missing list IDs.

sabri414 (dismissed):
Thanks for iterating on this! Reviewed the current diff and ran the targeted tests/lint plus a server-render smoke check for the hook. Everything looks good to me. Approving.

components/List/useKanbanPref.ts
```diff
@@ -17,20 +17,73 @@
...
+  const isKanban = kanbanPrefMap.get(listId) ?? false;
+
+  /**
+   * Sets the kanban preference in localStorage and updates the state
+   *
+   * @param value Value that the localStorage kanban preference will set to
+   */
+  const setIsKanban = useCallback(
+    (value: boolean | ((prev: boolean) => boolean)) => {
+      const kanbanPref = localStorage.getItem('KanbanPreference') ?? '[]';
```
radiantBear: For consistency, `const kanbanPref = getSnapshot()` would probably be better here

components/List/useKanbanPref.ts
```diff
@@ -17,20 +17,73 @@
...
+      const kanbanPrefMap = new Map<string, boolean>(kanbanPrefParsed);
+
+      if (!kanbanPrefMap.has(listId)) {
+        kanbanPrefMap.set(listId, false);
+        localStorage.setItem(
+          'KanbanPreference',
+          JSON.stringify(Array.from(kanbanPrefMap.entries()))
+        );
+        window.dispatchEvent(new Event('local-storage-update'));
+      }
```
radiantBear: Unless I'm missing something, this can be entirely removed. It just sets the list's preference to `false` immediately before setting it to the actual desired value. It's not needed on line 75, either, since it has the nullish coalesce to `false`.

radiantBear (dismissed):

sabri414 (dismissed):
This looks great!

radiantBear (approved):

sabri414 (approved):

PR #171, opened 2026-06-02
chore: update supply chain security vulns

Ryangman:
Updates supply chain introduced security holes. Does not update any packages via breaking changes, or packages without reported vulnerabilities.

### Review Notes
please perform a manual check that basic functionality works, particularly where not covered by e2e tests.


sabri414 (changes requested):
Requesting changes because this still leaves production audit failures. On the PR head, `npm audit --omit=dev` reports 3 moderate vulnerabilities: `better-auth@1.5.4` and `next@16.2.7` via nested `postcss@8.4.31`. Since this PR is specifically a supply-chain security cleanup, we should either resolve these or document an explicit exception before merging.

sabri414 (approved):
Approving, given the stated scope of avoiding breaking changes / problematic dependency resolution. I confirmed the remaining audit findings are not cleanly fixable right now: Next’s nested PostCSS issue requires an unsafe/breaking audit path, and updating Better Auth hits npm peer dependency conflicts. Lint and Vitest passed on the PR branch.

radiantBear (approved):

